PR #10, opened 2019-04-26
Using any of the available lattices at any given time, instead of a specifically named one

msohaibalam:
This helps avoid test failures anytime the available lattice(s) changes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pyquil>=2.4
+pyquil>=2.7
 pennylane>=0.2
 networkx
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ with open("pennylane_forest/_version.py") as f:
 
 
 requirements = [
-    "pyquil>=2.4",
+    "pyquil>=2.7",
     "pennylane>=0.2"
 ]
 
--- a/tests/test_qvm.py
+++ b/tests/test_qvm.py
@@ -14,8 +14,12 @@ from conftest import I, Z, H, U, U2, test_operation_map
 
 import pennylane_forest as plf
 
+import pyquil
+
 
 log = logging.getLogger(__name__)
+
+VALID_QPU_LATTICES = [qc for qc in pyquil.list_quantum_computers() if 'qvm' not in qc]
 
 
 class TestQVMBasic(BaseTest):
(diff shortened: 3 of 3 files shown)


msohaibalam:
@josh146 I might need your help on this one. It seems that there are a few tests that are failing because I can't seem to do the following:
```
>>> import pennylane as qml
>>> dev_numpy = qml.device('forest.numpy_wavefunction', wires=2)
```
which gives me an error message saying `pennylane._device.DeviceError: Device does not exist. Make sure the required plugin is installed.`

I could trace it back to https://github.com/XanaduAI/pennylane/blob/master/pennylane/__init__.py#L216, but don't exactly know how to resolve this issue.

josh146:
Hi @msohaibalam, I'm assuming this a test suite you are running locally, as I can't find the same error message in the Semaphore log? In order to have installed devices become available to PennyLane, the `setup.py install` command needs to be run on the plugin at some point - the best way is to install the plugin using pip in development mode:

```console
pip install -e .
```

Usually I get the above error if I forget to install the plugin (for example, by cloning the repository and then directly running pytest)

msohaibalam:
I thought I had already ran `pip install -e .` but apparently not! Thanks for the input, my tests (indeed, this was for a local test suite) are now passing. I'll have to dig into the error msgs in the Semaphore log a bit more to really get to the bottom of the failing build though.

msohaibalam:
@karalekas Semaphore is failing tests that pass locally for me. Most (all?) of them seem to be 
```
E       RuntimeError: Error encountered while parsing the quil program at line 1 and column 5
E       Received an '    ' but was expecting one of [ ' ', NEWLINE ]
```
Any ideas why this is happening?

josh146:
I think I've narrowed down the cause of the error. When I run the tests locally, I'm running them against pyquil version 2.7.2. However, I noticed that Semaphore is installing pyquil 2.6 due to caching; running the tests locally on 2.6 causes the same test errors.

I've pushed an update to the `setup.py` and `requirements.txt` to specify `pyquil>=2.7`, and see if that fixes the Semaphore build.

josh146:
Seems to have worked!

msohaibalam:
@josh146 Awesome! Merging this in now.

Issue #9, opened 2019-04-21
Have Pennylane-Forest use operator_estimation, particularly the measure_observables(..) function.


josh146:
> The QVM allows computing an expectation value with the `.expectation` call. The NumPy simulator lets you, within Python, inspect the wavefunction without it being transmitted. Generally, the NumPy simulator will be faster for small numbers of qubits or very rapid evaluations of many programs on a small number of qubits, since there's no interprocess communication overhead.
>
> See the [`expectation` method](https://github.com/rigetti/pyquil/blob/master/pyquil/api/_wavefunction_simulator.py#L77). This gets computed analytically.
>
> At some point we might also want to have Pennylane-Forest talk to operator_estimation, particularly the measure_observables(..) function.

_Originally posted by @stylewarning and @msohaibalam in https://github.com/rigetti/pennylane-forest/pull/4_

PR #13, opened 2019-05-18
Adds ability to return multi-qubit expectation values to all devi…

josh146:
In [PennyLane core](https://github.com/XanaduAI/PennyLane), we have added the ability to use `qml.expval.Hermitian(A, wires)` on an arbitrary number of wires. That is, you can now pass a matrix `A` of size `[2**N, 2**N]` which acts on `N` qubits (i.e., `len(wires)==N`).

This PR adds support to the PennyLane-Forest plugin for this feature. This includes the following changes:

* `forest.numpy_wavefunction`: replaces the `expand_one` method with the `expand` method, allowing an N qubit observable to be expanded to act on the entire system. Corresponding tests also added.

* `forest.wavefunction`: replaces the `expand_one` method with the `expand` method, allowing an N qubit observable to be expanded to act on the entire system. Corresponding tests also added.

* `forest.qvm`: modifies the `expval` method so that, if a multi-qubit observable is requested, the device calculates the wavefunction probablities, and uses this (alongside the eigenvalues of A) to reconstruct the required expectation value. Corresponding test `test_multi_mode_hermitian_expectation` also added.

* `forest.qpu`: inherits the above functionality from `forest.qvm`.

**Note:** This feature requires the latest development version of PennyLane.


pennylane_forest/numpy_wavefunction.py
```diff
@@ -90,25 +90,44 @@ def ev(self, A, wires):
...
+            # total number of wires is 1, simply return the matrix
+            return U
+
+        N = self.num_wires
+        wires = np.asarray(wires)
+
+        if np.any(wires < 0) or np.any(wires >= N) or len(set(wires)) != len(wires):
+            raise ValueError('Bad target subsystems.')
+
+        # generate N qubit basis states via the cartesian product
```
msohaibalam: Did you mean `tensor product` instead of `cartesian product`? (See e.g. https://physics.stackexchange.com/questions/350508/tensor-product-vs-cartesian-product-for-composite-quantum-systems)

pennylane_forest/numpy_wavefunction.py
```diff
@@ -90,25 +90,44 @@ def ev(self, A, wires):
...
-        U = np.kron(np.kron(np.eye(before), U), np.eye(after))
-        return U
+        if self.num_wires == 1:
+            # total number of wires is 1, simply return the matrix
+            return U
+
+        N = self.num_wires
+        wires = np.asarray(wires)
+
+        if np.any(wires < 0) or np.any(wires >= N) or len(set(wires)) != len(wires):
```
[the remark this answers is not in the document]
msohaibalam: It might also be worth checking that `U` has shape `(2**len(wires), 2**len(wires))`.

pennylane_forest/numpy_wavefunction.py
```diff
@@ -90,25 +90,44 @@ def ev(self, A, wires):
...
-    def expand_one(self, U, wires):
-        r"""Expand a one-qubit operator into a full system operator.
+    def expand(self, U, wires):
+        r"""Expand a multi-qubit operator into a full system operator.
 
         Args:
-          U (array): :math:`2\times 2` matrix
-          wires (Sequence[int]): target subsystem
+          U (array): :math:`2^n \times 2^n` matrix where n = wires.
+          wires (Sequence[int]): Target subsystems (order matters!)
```
msohaibalam: Maybe we should say a bit more about the order here. If I understood the functionality correctly, the left-most Hilbert space is at index 0.

pennylane_forest/qvm.py
```diff
@@ -170,21 +170,74 @@ def pre_expval(self):
...
+            # we know that p0+p1=1 and that <Z>=p0-p1
+            p0 = (1+evZ)/2
+            p1 = (1-evZ)/2
 
-        if expectation == 'Identity':
-            # <I> = \sum_i p_i
-            return p0+p1
+            if expectation == 'Identity':
+                # <I> = \sum_i p_i
+                return p0+p1
```
msohaibalam: Minor quibble, but this and other similar expressions might be more readable if there are spaces between the operands and binary operator, so e.g. `p0+p1` -> `p0 + p1` etc.

pennylane_forest/wavefunction.py
```diff
@@ -152,25 +152,44 @@ def ev(self, A, wires):
...
         """
         # Expand the Hermitian observable over the entire subsystem
-        A = self.expand_one(A, wires)
+        A = self.expand(A, wires)
         return np.vdot(self.state, A @ self.state).real
 
-    def expand_one(self, U, wires):
-        r"""Expand a one-qubit operator into a full system operator.
+    def expand(self, U, wires):
+        r"""Expand a multi-qubit operator into a full system operator.
```
msohaibalam: Most (if not all) of the comments I made for `numpy_wavefunction.py` apply here too.

tests/test_pyqvm.py
```diff
@@ -154,6 +154,36 @@ def test_hermitian_expectation(self, shots):
 
         self.assertAllAlmostEqual(res, expected, delta=3/np.sqrt(shots))
 
+    def test_multi_mode_hermitian_expectation(self, shots, qvm, compiler):
+        """Test that arbitrary multi-mode Hermitian expectation values are correct"""
+        theta = 0.432
+        phi = 0.123
```
msohaibalam: Instead of picking hard values like this, how about picking from some range at random and then fixing the seed? This way, it's obvious from the start that `theta` and `phi` are parameters of a more generally applicable expression.

pennylane_forest/numpy_wavefunction.py
```diff
@@ -90,25 +90,44 @@ def ev(self, A, wires):
           float: expectation value :math:`\left\langle{A}\right\rangle = \left\langle{\psi}\mid A\mid{\psi}\right\rangle`
         """
         # Expand the Hermitian observable over the entire subsystem
-        A = self.expand_one(A, wires)
+        A = self.expand(A, wires)
         return np.vdot(self.state, A @ self.state).real
 
-    def expand_one(self, U, wires):
```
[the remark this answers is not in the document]
josh146: This method is an internal method, however --- it is not (or at least, there is no case where it should be) user or developer accessible, so a deprecation notice seems redundant. The new `expand` method should probably be renamed `_expand` instead

pennylane_forest/numpy_wavefunction.py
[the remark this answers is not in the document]
josh146: Added these in!

pennylane_forest/numpy_wavefunction.py
josh146: Done

pennylane_forest/numpy_wavefunction.py
josh146: Ah, I meant it more in the programmatic sense, in that `itertools.product` performs the cartesian product

pennylane_forest/numpy_wavefunction.py
```diff
@@ -90,25 +90,44 @@ def ev(self, A, wires):
...
+
+        N = self.num_wires
+        wires = np.asarray(wires)
+
+        if np.any(wires < 0) or np.any(wires >= N) or len(set(wires)) != len(wires):
+            raise ValueError('Bad target subsystems.')
+
+        # generate N qubit basis states via the cartesian product
+        tuples = np.stack(np.meshgrid(*[[0, 1] for _ in range(N)]), -1).reshape(-1, N)
+        tuples[:, [0, 1]] = tuples[:, [1, 0]]
```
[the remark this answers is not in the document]
josh146: I've made the replacement :slightly_smiling_face: 

I was trying to be clever, assuming the NumPy vectorised approach would be faster, but this is not the case:
```python
In [17]: %timeit np.array(list(product([0, 1], repeat=4)))
9.87 µs ± 347 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [18]: %timeit np.stack(np.meshgrid(*[[0, 1] for _ in range(4)]), -1).reshape(-1, 4)
62.5 µs ± 1.38 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

pennylane_forest/numpy_wavefunction.py
```diff
@@ -90,25 +90,44 @@ def ev(self, A, wires):
...
+
+        # wires not acted on by the operator
+        inactive_wires = list(set(range(N))-set(wires))
+
+        # expand U to act on the entire system
+        U = np.kron(U, np.identity(2**len(inactive_wires)))
+
+        # move active wires to beginning of the list of wires
+        rearanged_wires = np.array(list(wires)+inactive_wires)
```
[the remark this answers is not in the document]
josh146: oops! good catch

pennylane_forest/qvm.py
josh146: You're right, this is also specified in PEP8 :+1:

pennylane_forest/wavefunction.py
josh146: Copied over the fixed function :slightly_smiling_face:

tests/test_pyqvm.py
josh146: Done :+1:

josh146:
Thanks @msohaibalam for your review, I've just pushed a commit with all suggested changes.

The tests are likely going to fail for two reasons:

1. There have been breaking changes in pyQuil 2.8, I'll make a new PR now to address this

2. This PR requires PennyLane GitHub master installed for these two tests to pass --- semaphore is simply installing the pip version. We are working on updating the version on pip this week.

josh146:
Note: this is likely a stop-gap solution, once #9 is implemented, it will likely be faster and more efficient than the matrix expansion done here.

pennylane_forest/numpy_wavefunction.py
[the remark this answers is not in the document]
msohaibalam: Got it. Doesn't need to happen in this PR, but at some point, it'd be a good idea to add leading underscores to all functions that are really meant to be private, just to avoid confusion.

pennylane_forest/numpy_wavefunction.py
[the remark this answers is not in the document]
msohaibalam: That's a cool little factoid! I wouldn't have guessed the faster run time either, and instead was purely arguing from a readability perspective.

Should we add `itertools` as part of the requirements?

pennylane_forest/numpy_wavefunction.py
msohaibalam: Ok, that makes sense.

pennylane_forest/qvm.py
msohaibalam: There are a few other spots where we could make a similar change: `p0 = (1+evZ)/2` and `p1 = (1-evZ)/2` (lines 179-180), `return w[0]*p0 + w[1]*p1` -> `return w[0] * p0 + w[1] * p1` (line 190), `cb = np.sum(2**np.arange(len(wires)-1, -1, -1)*i)` -> `cb = np.sum(2**np.arange(len(wires)-1, -1, -1) * i)` (line 233) all in `qvm.py`.

In `numpy_wavefunction.py`: `ev = (n0*a[0] +(self.shots-n0)*a[1]) / self.shots` -> `ev = (n0 * a[0] +(self.shots - n0) * a[1]) / self.shots` (line 80), `inactive_wires = list(set(range(N))-set(wires))` -> `inactive_wires = list(set(range(N)) - set(wires))` (line 126), `rearranged_wires = np.array(list(wires)+inactive_wires)` -> `rearranged_wires = np.array(list(wires) + inactive_wires)` (line 132), `perm = np.ravel_multi_index(tuples[:, rearranged_wires].T, [2]*N)` -> `perm = np.ravel_multi_index(tuples[:, rearranged_wires].T, [2] * N)` (line 138).

josh146: What I might do is apply the [black](https://github.com/python/black) Python formatter to all source codes - I'll do that in a separate PR, so as not to mess with the diff too much.

msohaibalam: Sounds good to me!

tests/test_pyqvm.py
msohaibalam: Why not do this for all the `theta = 0.432`, `phi = 0.123` definitions in the tests in `test_pyqvm.py`?

msohaibalam: There are also similar (but different) parameters in `test_one_qubit_wavefunction_circuit` that we could probably similarly replace with randomly picked values.

josh146: Ah, I didn't spot the others in the diff. Does it make sense to do this in a new PR?

msohaibalam: Sure, that sounds good to me.

PR #16, opened 2019-06-27
Adds variance support

josh146:
This PR adds support for the `pennylane.var()` measurement return type newly added to PennyLane v0.4.

* Adds the device method `var()` to all devices

* Adds corresponding tests, to ensure the variance is correctly calculated.

* Updates the minimum required PennyLane version to 0.4

* Update the documentation to reflect PennyLane v0.4 support


josh146:
Thanks @msohaibalam! I'll create a PR to run [Black](https://github.com/python/black) for PEP8 formatting on the entire module.

PR #24, opened 2019-10-18
Updates the plugin to work with PyQuil 2.13 and PennyLane 0.6

josh146:
**Description of changes**:

* Version bump to 0.6.0, updated requirements and setup.py

* The way measurement statistics works has changed in the latest version of PennyLane. Now, rather than `shots=0` referring to 'analytic' mode, there is a separate `analytic` argument. Further, the `num_shots` argument has been removed from `Device.samples()`.

* Tests that are no redundant due to the above change have been removed.

**Drawbacks**:

Note that there are two outstanding issues related to versions of PyQuil >= 2.11:

* PyQVM is significantly slower compared to PyQuil 2.10

* `QubitUnitary` no longer works with the NumPy Wavefunction Simulator (see https://github.com/rigetti/pyquil/issues/1059).

Once merged, a new release should be made to PyPI.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,7 +71,7 @@ We also encourage bug reports, suggestions for new features and enhancements, an
 Authors
 =======
 
-`Josh Izaac <https://github.com/josh146>`_, `Keri A. McKiernan <https://github.com/kmckiern>`_
+`Josh Izaac <https://github.com/josh146>`_, `Keri A. McKiernan <https://github.com/kmckiern>`_, `M. Sohaib Alam <https://github.com/msohaibalam>`_
 
 
 Support
--- a/README.rst
+++ b/README.rst
@@ -9,9 +9,9 @@ PennyLane Forest Plugin
     :alt: Documentation Status
     :target: http://pennylane-forest.readthedocs.io/en/latest/?badge=latest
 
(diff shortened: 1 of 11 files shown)


josh146:
The test that fails (`test_sample_values_hermitian`) is a notoriously stochastic integration test 🙂 Re-triggering the build should fix it. The tolerance on that test is already quite high, and I'm not sure of the best way to make it 'more deterministic'?

msohaibalam:
It is pretty odd since the tolerance factors seem to respect calculated standard errors of the mean. Locally, I also see more frequent passes than fails. And yet on Semaphore I've seen only fails so far with the rebuilds. If all else fails, maybe just increase the no. of shots even more, or fix a random seed? I don't really have a better suggestion than that I'm afraid.

msohaibalam:
@josh146 I created a new branch with a much greater number of shots, but that too seems to be failing on the same test. I'm pretty convinced though that this is not an issue with the functionality itself, and am totally ok with merging this in to master. That will also unblock other PRs that need to start making their way in now.

josh146:
Hmm this is really odd. As far as I can tell, the stochasticity from sampling should not lead to such frequent failures as we are seeing...

But yes, if it can be merged, we can look into it separately!

msohaibalam:
Looks like I finally have passing tests! Created PR #27 to merge into this branch, so we'll end up with passing tests here and then master.

josh146:
Thanks for getting this passing @msohaibalam!

Issue #20, opened 2019-08-12
Illegal qubits when using QPU


josh146:
When using the QVM to simulate hardware devices such as Aspen-4-9Q-A, i.e.,

```python
qml.device("forest.qvm", device="Aspen-4-9Q-A")
```

PennyLane-Forest works as expected. From the main interface, PL wires (labeled 0-8) are converted to match the topology of the QPU (labelled 0,1,2,7,10,11,15,16,17).

However, when attempting to run the exact same program on the QPU,

```python
qml.device("forest.qpu", device="Aspen-4-9Q-A")
```

it produces the error
```python
Error: "Program instruction RX(pi/4) attempts to use illegal qubits: 3. Illegal qubits on this QPU: 3, 4, 5, 6, 8, 9, 12, 13"
```

Is there an error in how PennyLane-Forest perform the 'rewiring' of qubits for the QPU? Ideally, a computation with qubits labelled 0-8 should run seamlessly and be rewired behind the scenes for the user.

EDIT/UPDATE: There are two related issues here: (i) the one outlined above, and (ii) the in-ability to run programs on legal qubits on the lattice. Issue (i) can be illustrated by the following code snippet
```
import pennylane as qml
from pennylane import numpy as np

qpu = qml.device("forest.qpu", device="Aspen-4-4Q-E")
print(qpu.qc.qubits())  # [10, 11, 12, 17]

@qml.qnode(qpu)
def circuit(angle):
    qml.RY(angle, wires=0)
    return qml.expval(qml.PauliZ(0))

circuit(np.pi/2)   # raises `RPCError: Program instruction 'RY(pi/2) 0' attempts to use illegal qubits: 0. Illegal qubits on this QPU: 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 13, 14, 15, 16.`
```

while issue (ii) can be illustrated by the following code snippet
```
import pennylane as qml
from pennylane import numpy as np

qpu = qml.device("forest.qpu", device="Aspen-4-4Q-E")
print(qpu.qc.qubits())  # [10, 11, 12, 17]

@qml.qnode(qpu)
def circuit(angle):
    qml.RY(angle, wires=10)
    return qml.expval(qml.PauliZ(10))

circuit(np.pi/2)    # raises `QuantumFunctionError: Operation RY applied to invalid wire 10 on device with 4 wires.`
```

msohaibalam:
Hi @josh146 , good to play around with pennylane-forest code after my break!
[...]

PR #25, opened 2019-10-31
QPU Parity

msohaibalam:
Addressing issue #20 and increasing parity between the QVM and QPU abstractions. Circuits run on e.g. wire 0 will now run on the (0-indexed) first qubit on the physical lattice.

### DIFF
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -102,7 +102,7 @@ class QPUDevice(QVMDevice):
         self.active_reset = active_reset
         self.symmetrize_readout = symmetrize_readout
         self.calibrate_readout = calibrate_readout
-        self.wiring = {q: i for i, q in enumerate(self.qc.qubits())}
+        self.wiring = {i: q for i, q in enumerate(self.qc.qubits())}
 
     def pre_rotations(self, observable, wires):
         """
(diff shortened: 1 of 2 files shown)


msohaibalam:
@josh146 I was initially planning to give the user an option to "reverse wire" as well. But on 2nd thought, since it seems that Pennylane's default and intended use is to use the 0..n_qubits-1 indexing convention, it might make more sense to just force the "wiring" introduced in this PR, and not allow the user to define their own. Open to suggestions/thoughts on this however.

pennylane_forest/qpu.py
```diff
@@ -111,21 +112,27 @@ def pre_rotations(self, observable, wires):
...
-            "Hadamard": [ExperimentSetting(TensorProductState(), float(np.sqrt(1/2)) * sX(0)),
-                         ExperimentSetting(TensorProductState(), float(np.sqrt(1/2)) * sZ(0))]
-        }
-        # expectation values for single-qubit observables
+        # Single-qubit observable
         if len(wires) == 1:
-
+            # identify Experiment Settings for each of the possible single-qubit observables
+            wire = wires[0]
+            qubit = self.wiring[wire]
```
josh146: Ends up being a very simple fix! 🙂

msohaibalam:
> So for instance,
> 
> ```python
> dev = qml.device("forest.qpu", backend="Aspen-4", wire_numbering="physical")
> ```
> 
> which allows the user to use the physical qubit ordering?

Yes, that's exactly what I had in mind. Although I now realize [this check](https://github.com/XanaduAI/pennylane/blob/master/pennylane/qnode.py#L675) in `pennylane.qnode` will block us from doing this, and will raise an error on trying to run a circuit on e.g. wire (qubit) 10 on "Aspen-4-4Q-E". So maybe table this for now?

> I would like to, when we have time, refactor this plugin to better match PyQuil terminology and 
> UX. I.e., merge forest.qvm with forest.qpu, maybe change the device="Aspen-*" keyword 
> argument to qc="Aspen-*.

I agree, we should work this out at some point. For now, I've created an issue to address it: https://github.com/rigetti/pennylane-forest/issues/28.

msohaibalam:
@josh146 As we discussed offline, let's first work on updating PennyLane to handle the "physical" wiring of the QPUs, and then produce that capability within PennyLane-Forest. Merging these changes into master.

PR #36, opened 2019-12-24
Assuming 1q and 2q gates in QPU.expval

### LINKED ISSUE #34: QNodes with 2-qubit gates fail on the QPU
The new operator estimation method in the QPU class currently works for single qubit gates, however fails as soon as a multi-qubit gates are included within the QNode. The traceback is as follows:

```text
~/pennylane-forest/pennylane_forest/qpu.py in expval(self, observable, wires, par)
    132                     if isinstance(instr, Gate):
    133                         # assumes single qubit gates
--> 134                         gate, _ = instr.out().split(' ')
    135                         prep_prog += Program(str(gate) + ' ' + str(qubit))
    136                 if self.readout_error is not None:
ValueError: too many values to unpack (expected 2)
```

msohaibalam:
The preparation prog for operator estimation can now handle 1q and 2q gates (but note that it probably won't work with statement involving more than one qubit, e.g. those involving `CONTROLLED` etc).

Fixes #34 .

### DIFF
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -130,9 +130,15 @@ class QPUDevice(QVMDevice):
                 prep_prog = Program()
                 for instr in self.program.instructions:
                     if isinstance(instr, Gate):
-                        # assumes single qubit gates
-                        gate, _ = instr.out().split(' ')
-                        prep_prog += Program(str(gate) + ' ' + str(qubit))
+                        # split gate and wires -- assumes 1q and 2q gates
+                        tup_gate_wires = instr.out().split(' ')
+                        gate = tup_gate_wires[0]
+                        str_instr = str(gate)
+                        # map wires to qubits
+                        for w in tup_gate_wires[1:]:
+                            str_instr += f' {self.wiring[int(w)]}'
+                        prep_prog += Program(str_instr)
+
                 if self.readout_error is not None:
                     prep_prog.define_noisy_readout(qubit, p00=self.readout_error[0],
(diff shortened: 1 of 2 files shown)


msohaibalam:
@kilimanjaro Hey Erik, pinging you so you could get eyes on this as well.
@josh146 ^ Erik D's github handle, fyi.

trbromley:
Thanks @msohaibalam, just gave it a try on the QPU and it works without any error.
:+1:

msohaibalam:
Great to hear @trbromley! Glad at least that bug is resolved. However, I am now concerned about the failing tests I'm seeing (both on Semaphore and locally) where the compiler simply stops running.

josh146:
@msohaibalam, I can see this locally as well :thinking:

msohaibalam:
@josh146 I can get these tests to pass locally now, with the updated `quilc`. I got a failing test because of statistical fluctuations, performed a rebuild, and got a timeout error (though different from the one I was seeing with quilc). Going to try another rebuild or two before re-thinking my strategy.

josh146:
Okay good to know! After the update to quilc, we are currently noticing another odd bug in pyQuil/Quilc, where the compiler will timeout if we have > 2 qubits and a multi-qubit gate present --- is that the same thing happening here?

msohaibalam:
Strange, I haven't seen that bug on my side. The tests now pass on semaphore, so I'll merge this in. If the bug you mentioned persists, we can work on that in a new PR (my guess is it would be non-PLF related anyway).

PR #37, opened 2020-01-13
Fixes the plugin to support pyQuil 2.16

josh146:
This PR updates the PennyLane-Forest plugin to support breaking API changes in pyQuil 2.16. In particular:

* The import `pyquil.gate_matrices.QUANTUM_GATES` no longer works, and has been changed to `_matrix_dictionary = pyquil.simulation.matrices.QUANTUM_GATES`.

* `pyquil.numpy_simulator` has been changed to `pyquil.NumpyWavefunctionSimulator`

* `TomographyExperiment` has been changed to `Experiment`.

There are still quite a few deprecation notices, specifically related to tomography:

* FutureWarning: `n_shots` has been deprecated; if you want to set the number of shots for this run of measure_observables please provide the number to `Program.wrap_in_numshots_loop()` for the Quil program that you provide when creating your `TomographyExperiment` object. It looks like your `TomographyExperiment` object has `shots = 1`, so for now we will change that to 10000, which was the previous default value.

* FutureWarning: `active_reset` has been deprecated; if you want to enable active qubit reset please provide a Quil program that has a `RESET` instruction in it when creating your `TomographyExperiment` object. For now, this value will override that in the `TomographyExperiment`, but eventually this keyword argument will be removed.

* FutureWarning: `symmetrize_readout` has been deprecated; please provide the symmetrization level when creating your `TomographyExperiment` object. For now, this value will override that in the `TomographyExperiment`, but eventually this keyword argument will be removed.

@msohaibalam, I'm a bit unsure the best approach to fix these three (I assume the only high priority one is `n_shots`?). Furthermore, is `active_reset` deprecated only for tomography, or for generic QPU runs as well? We do not get a deprecation notice there.

Finally, there is an additional warning I have not noticed before:

```
UserWarning: SIMPLE-WARNING: Chip specification contained fidelity 1.0d0 > 0.99999d0. Truncating to 0.99999d0
```

### DIFF
--- a/pennylane_forest/converter.py
+++ b/pennylane_forest/converter.py
@@ -324,7 +324,7 @@ class ProgramLoader:
         program (pyquil.quil.Program): The pyquil Program instance that should be loaded
     """
 
-    _matrix_dictionary = pyquil.gate_matrices.QUANTUM_GATES
+    _matrix_dictionary = pyquil.simulation.matrices.QUANTUM_GATES
 
     def __init__(self, program):
         self.program = program
--- a/pennylane_forest/device.py
+++ b/pennylane_forest/device.py
@@ -226,7 +226,12 @@ class ForestDevice(Device):
 
     def apply(self, operation, wires, par):
         # pylint: disable=attribute-defined-outside-init
-        self.prog += self._operation_map[operation](*par, *[int(w) for w in wires])
+        if hasattr(self, "wiring"):
+            qubits = [int(self.wiring[i]) for i in wires]
+        else:
+            qubits = [int(w) for w in wires]
+
+        self.prog += self._operation_map[operation](*par, *qubits)
 
         # keep track of the active wires. This is required, as the
         # pyQuil wavefunction simulator creates qubits dynamically.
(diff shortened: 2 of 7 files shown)


trbromley:
I have just checked running on QPU and this also fixes the error we were seeing there. Thanks @josh146 !

pennylane_forest/converter.py
```diff
@@ -324,7 +324,7 @@ class ProgramLoader:
         program (pyquil.quil.Program): The pyquil Program instance that should be loaded
     """
 
-    _matrix_dictionary = pyquil.gate_matrices.QUANTUM_GATES
+    _matrix_dictionary = pyquil.simulation.matrices.QUANTUM_GATES
```
kilimanjaro: Would suggest bumping to `pyquil>=2.7` in `requirements.txt`

josh146: Ah, good catch!

josh146:
@kilimanjaro, just realised I can't merge this as you don't have write access to the repo (so github doesn't count your approval)!

Issue #7, opened 2019-03-07
How come you don't use memory_map dict as qc.run() argument?


rrtucci:
According to the Forest docs, the "new way" of minimizing an anzatz does the following:
1. declares parameters in the program, and
2. uses the memory_map dict as an argument to qc.run(). 

But your classes
 device, ForestDevice, qvm, qpu
none of them does this. Is this then the "old way" of minimizing the anzatz? If so, why are you using the old way?

josh146:
Hi @rrtucci, you raise a good point. PennyLane and the Forest plugin were developed right before parametric compilation was supported in pyQuil (using pyQuil v1.9), so at the time, the plugin was designed around the 'old' way.

It would be great to also support pyQuil v2 style parametric programs.

One complication is that PennyLane takes care of the symbolic variables automatically, and passes the resulting numeric values to the plugin device when an evaluation is needed. This just means that the core PennyLane library will also need a pull request to have the core library pass which parameters are symbolic to the plugin.

josh146:
@msohaibalam, any thoughts?

rrtucci:
Thanks, Josh. I am learning a tremendous amount from PennyLane. You are a great programmer

msohaibalam:
Thanks for bringing attention to this! The important point is that your parametric program gets compiled once in order to avoid compilation overheads, so that you can scan over several parameters without re-compiling.

@josh146 Correct me if I'm wrong, but in the current set-up, we're limited to doing operations like the following:

```
import pennylane as qml
from pennylane import numpy as np


qvm = qml.device('forest.qvm', device='2q-qvm', wires=2, shots=1000)

@qml.qnode(qvm)
def circuit(theta):
    qml.RX(theta, wires=0)
    return qml.expval.PauliZ(0)

print(circuit(np.pi/2))
```

And if we do something like 
```
import pennylane as qml
from pennylane import numpy as np


qvm = qml.device('forest.qvm', device='2q-qvm', wires=2, shots=1000)

@qml.qnode(qvm)
def circuit():
    for theta in np.linspace(0.0, np.pi, 100):
        qml.RX(theta, wires=0)
        yield qml.expval.PauliZ(0)

print(list(circuit()))
```
we'll get the error msg: `QuantumFunctionError: A quantum function must return either a single expectation value or a nonempty sequence of expectation values.`

It would make more sense to use and support memory_map in qc.run() for something like the latter example.

josh146:
[...]

PR #39, opened 2020-01-28
Adding parametric compilation for QVM and QPU

antalszava:
**Context:**
[Parametric compilation](http://docs.rigetti.com/en/stable/basics.html#parametric-compilation) has been added as a feature to PyQuil a while ago. With this feature, certain circuits that contain symbolic parameters can be compiled once, allowing such parameters to be bound right before the program is run.

When running certain tests, some timeouts were experienced.

*(Note: this branch was branching off from the ``qubit_device`` branch, see the changes [in this PR](https://github.com/XanaduAI/pennylane/pull/483))*

1) Adds parametric compilation capabilities to ``QVMDevice`` and ``QPUDevice``.
2) Adds the keyword argument ``timeout`` to the ``QVMDevice`` or the ``QPUDevice``.

Upon turning **parametric compilation** on, a parametric program is being built. This program is built using the following instance attributes there were introduced for both previously mentioned devices:
*  *_lookup_table*: a dictionary that associates a ``QNode`` hash with a compiled program. Upon compilation, a new entry is added, should the program not have been compiled once already.
* *_parameter_map*: a dictionary that associates each symbolic variable in the circuit with it's numeric value for the specific evaluation of the circuit. This dictionary is built based on the ``Variable`` instances included in the specific ``QNode``.
* *_parameter_reference_map*: a dictionary that associates each symbolic variable in the circuit with it's PyQuil reference that has been created during the creation of the parametric program. The values of this dictionary are memory references as declared in the ``theta_ref = p.declare("theta", "REAL")`` fashion.

The flow of logic for building the parametric program is the following (``QVMDevice.apply_parametric_program``):
1) Each of ``operation`` is taken from the set of ``operations``
2) For each ``Variable`` found in ``operation.params``, a ``PyQuil`` parameter reference is created
3) The value of the parameter is stored in ``_parameter_map``
4) The parameter reference is stored in ``_parameter_reference_map``
5) Once the each operation has been processed, we proceed with the observables

Eventually, when the parametric program is to be run, the ``QuantumComputer.run`` is called by passing the ``memory_map=self._parameter_map`` keyword argument for substituting the numeric values.

**Benefits:**
Now the ``QVMDevice`` and ``QPUDevice`` have ``parametric_compilation=True`` by default. This means, that identical circuits defined in a ``QNode`` are only compiled *once*, regardless of the number of times they are called.

A ``timeout`` can be set when using the ``QVMDevice`` or the ``QPUDevice``.

**Possible Drawbacks:**

Certain attributes are currently defined in both the ``QVMDevice`` and ``QPUDevice``.

In a next PR, the QPU and QVM devices could be refactored, such that the QPU device cannot be called for as_qvm=True. Only the QVMDevice can be used like that.

The goal of this PR would further be to redesign the attributes of the ``QVMDevice`` and ``QPUDevice``, e.g. refactoring the ``super(QVMDevice, self).__init__(num_wires, shots, **kwargs)`` call in the __init__ method of ``QPUDevice`` and further reorganize the shared attributes for the two devices.

**Related GitHub Issues:**
N/A


pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
-        # the resulting probability of each computational basis state
-        probs = np.zeros([2 ** len(wires), 2])
-        probs[:, 0] = np.arange(2 ** len(wires))
+    def generate_samples(self):
+        if "pyqvm" in self.qc.name:
+            self._samples = self.qc.run(self.prog, memory_map=self._parameter_map)
+        else:
+            # No hash provided or parametric compilation was set to False
+            # Will compile the program
+            if self.circuit_hash is None or not self.parametric_compilation:
```
kilimanjaro: Where is `circuit_hash` getting set?

pennylane_forest/qpu.py
```diff
@@ -82,6 +82,14 @@ def __init__(self, device, *, shots=1024, active_reset=True, load_qc=True, reado
         self.readout_error = readout_error
 
         self._eigs = {}
+        self.parametric_compilation = kwargs.get("parametric_compilation", True)
+
+        if self.parametric_compilation:
+            self._lookup_table = {}
```
kilimanjaro: It's not clear to me where `_lookup_table` is used. And unlike the other two, it does not seem to be defined in `device.py`

pennylane_forest/qvm.py
antalszava: Hi, yes, that's a completely valid question, as based on only this PR it comes as something out of context.

Following up on the previous PR #38 , now ``ForestDevice`` inherits from ``QubitDevice``. The PennyLane type of flow of logic involves ``evaluating`` a quantum circuit by calling on the ``execute`` method of the device (such is ``QVMDevice``).

Each instance of the ``ForestDevice`` (now ``QVMDevice`` in our case) will currently inherit the same ``execute`` method [as seen here](https://github.com/XanaduAI/pennylane/blob/a1f22b10ee968ade10488c6c42a7b153553cf628/pennylane/_qubit_device.py#L123) (which is not redefined in any of the subclasses).

After some checks, the first statement of this method is 
``self._circuit_hash = circuit.hash``.

Here, ``circuit.hash`` is a hash created from the circuit for the python session [as seen here](https://github.com/XanaduAI/pennylane/blob/a1f22b10ee968ade10488c6c42a7b153553cf628/pennylane/circuit_graph.py#L187) and [here](https://github.com/XanaduAI/pennylane/blob/a1f22b10ee968ade10488c6c42a7b153553cf628/pennylane/circuit_graph.py#L140). Identical circuits will produce identical hashes. When it comes to comparing if parametric gates were called with the arguments, [numeric parameters are checked by value](https://github.com/XanaduAI/pennylane/blob/a1f22b10ee968ade10488c6c42a7b153553cf628/pennylane/circuit_graph.py#L165), [symbolic parameters are checked by reference](https://github.com/XanaduAI/pennylane/blob/a1f22b10ee968ade10488c6c42a7b153553cf628/pennylane/circuit_graph.py#L158). (In the implementation itself, a string is being built).

pennylane_forest/qpu.py
antalszava: There are two subtle things to consider here:
*  ``QPUDevice`` inherits the ``generate_samples`` method from ``QVMDevice``. This method uses ``_lookup_table`` to store compiled programs.
* Currently, ``QPUDevice__init__`` calls on the ``ForestDevice.__init__`` method and ``QVMDevice.__init__`` is not called, (coming from  [``super(QVMDevice, self).__init__(num_wires, shots, **kwargs)``](https://github.com/rigetti/pennylane-forest/blob/9f104463d4a101d09082952f1377236b0759da6e/pennylane_forest/qpu.py#L103)). As both the ``QPUDevice`` and ``QVMDevice``, require the dictionaries, they initialize them separately. This design and how the ``__init__`` methods are called could be altered in a subsequent PR.

As for the method ``generate_samples``, this is called in ``QubitDevice.execute`` (and thus for every device defined in this plugin).

pennylane_forest/qvm.py
antalszava: Just to sketch the flow of logic in ``execute``:
1) The circuit hash is set
2) The circuit operations and rotations into the measurement basis are applied (in our case this means building up the PyQuil ``Program``)
3) Samples are generated in the computational basis (in our case this involves compiling (if applicable) and running the ``Program``)
4) Statistics are extracted based on the samples obtained (e.g. returning expectation values, variances, etc.)

pennylane_forest/device.py
```diff
@@ -224,31 +224,53 @@ def program(self):
...
-        else:
-            self.active_wires = set(range(self.num_wires))
+    def apply_rotations(self, rotations):
+        """Apply the circuit rotations.
 
-    @abc.abstractmethod
-    def pre_measure(self):  # pragma no cover
-        """Run the QVM or QPU"""
-        raise NotImplementedError
+        This method serves as an auxiliary method to :meth:`~.ForestDevice.apply`.
```
josh146: It might also be good to mention in the docstring what this method does/how it changes the quantum program

pennylane_forest/device.py
```diff
@@ -224,31 +224,53 @@ def program(self):
...
+            self.prog += self._operation_map[operation.name](*par, *wires)
 
     def reset(self):
         self.prog = Program()
-        self.active_wires = set()
-        self.state = None
+        self._active_wires = set()
+        self._state = None
+        self._parameter_map = {}
+        self._parameter_reference_map = {}
```
josh146: What's the difference between the parameter map and parameter reference map?

pennylane_forest/qvm.py
```diff
@@ -71,13 +73,24 @@ class QVMDevice(ForestDevice):
...
     def __init__(self, device, *, shots=1024, noisy=False, **kwargs):
-        self._eigs = {}
 
         if shots <= 0:
             raise ValueError("Number of shots must be a positive integer.")
 
         # ignore any 'wires' keyword argument passed to the device
         kwargs.pop("wires", None)
+        analytic = kwargs.get("analytic", False)
+        timeout = kwargs.pop("timeout", None)
```
josh146: `timeout` should be added to the `Keyword Args:` section of the docstring 🙂

pennylane_forest/qvm.py
```diff
@@ -110,53 +123,19 @@ def __init__(self, device, *, shots=1024, noisy=False, **kwargs):
         elif isinstance(device, str):
             self.qc = get_qc(device, as_qvm=True, noisy=noisy, connection=self.connection)
 
+        if timeout:
```
josh146: The default value of `timeout` is `None`, so this should probably be
```suggestion
        if timeout is not None:
```

pennylane_forest/qvm.py
```diff
@@ -71,13 +73,24 @@ class QVMDevice(ForestDevice):
...
         kwargs.pop("wires", None)
+        analytic = kwargs.get("analytic", False)
+        timeout = kwargs.pop("timeout", None)
+
+        self.parametric_compilation = kwargs.get("parametric_compilation", True)
+
+        if self.parametric_compilation:
+            self._lookup_table = {}
+            self._parameter_map = {}
+            self._parameter_reference_map = {}
```
josh146: It would be nice to have attribute docstrings describing each attribute, as well as the type of the keys/values in the dictionary

pennylane_forest/qvm.py
josh146: But this is quite nice to now be able to set!

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
+    def apply_parametric_program(self, operations, **kwargs):
+        # pylint: disable=attribute-defined-outside-init
+        rotations = kwargs.get("rotations", [])
+
+        # Storing the active wires
+        self._active_wires = ForestDevice.active_wires(operations + rotations)
+
+        # Apply the circuit operations
+        for i, operation in enumerate(operations):
+            # number of wires on device
```
josh146: Rather than the number of wires on the device, should this instead say the following?
```suggestion
            # map the operation wires to the physical device qubits
```

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
+            if i > 0 and operation.name in ("QubitStateVector", "BasisState"):
+                raise DeviceError("Operation {} cannot be used after other Operations have already been applied "
+                                  "on a {} device.".format(operation.name, self.short_name))
+
+            # Prepare for parametric compilation
+            par = []
+            for param in operation.params:
+                if isinstance(param, Variable):
+                    # Using the idx for each Variable instance
+                    parameter_string = "theta" + str(param.idx)
```
josh146: Nice solution!

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
+                    if parameter_string not in self._parameter_map:
+                        current_ref = self.prog.declare(parameter_string, "REAL")
+                        self._parameter_reference_map[parameter_string] = current_ref
```
josh146: It might be good to have line comments here, but if my understanding is correct:

1. We check if the parameter string has been used before(?)

2. If not, a Quil parameter is created (with type real), and added to the parameter reference map

3. Finally, the parameter map maps the parameter string to the parameter value

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
+
+                    self._parameter_map[parameter_string] = [param.val]
```
josh146: How come this is a list on the right hand side?

pennylane_forest/qvm.py
```diff
@@ -71,13 +73,24 @@ class QVMDevice(ForestDevice):
...
 
         if shots <= 0:
             raise ValueError("Number of shots must be a positive integer.")
 
         # ignore any 'wires' keyword argument passed to the device
         kwargs.pop("wires", None)
+        analytic = kwargs.get("analytic", False)
+        timeout = kwargs.pop("timeout", None)
+
+        self.parametric_compilation = kwargs.get("parametric_compilation", True)
```
josh146: This should also be added to the docstring :)

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
-        res = np.array([self.state[w] for w in wires]).T
-        for i in res:
-            # for each sample, calculate which
-            # computational basis state it corresponds to
-            cb = np.sum(2 ** np.arange(len(wires) - 1, -1, -1) * i)
-            # add a tally for this computational basis state
-            # to our array of basis probabilities
-            probs[cb, 1] += 1
+            # Store the compiled program with the corresponding hash
+            elif self.circuit_hash not in self._lookup_table:
```
josh146: Maybe it might be better to give `self._lookup_table` a more descriptive name --- how about `self._compiled_program_dict`?

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
-        # create an array of size [2^len(wires), 2] to store
-        # the resulting probability of each computational basis state
-        probs = np.zeros([2 ** len(wires), 2])
-        probs[:, 0] = np.arange(2 ** len(wires))
+    def generate_samples(self):
+        if "pyqvm" in self.qc.name:
+            self._samples = self.qc.run(self.prog, memory_map=self._parameter_map)
+        else:
+            # No hash provided or parametric compilation was set to False
+            # Will compile the program
```
josh146: Minor comment, but it might be easier to parse if the comments are placed after each branch, e.g.,
```python
if self.circuit_hash is None or not self.parametric_compilation:
    # No hash provided or parametric compilation was set to False
    # Will compile the program
```
(same follows for the elif and else)

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
+                compiled_program = self.qc.compile(self.prog)
+                self._lookup_table[self.circuit_hash] = compiled_program
 
-        # sort the probabilities by the first column,
-        # and divide by the number of shots
-        probs = probs[probs[:, 0].argsort()] / self.shots
-        probs = probs[:, 1]
+            # The program has been compiled already
+            else:
+                compiled_program = self._lookup_table[self.circuit_hash]
```
josh146: Awesome!

pennylane_forest/qvm.py
```diff
@@ -165,80 +144,66 @@ def pre_measure(self):
...
 
-        return probs
+            self._samples = self.qc.run(executable=compiled_program, memory_map=self._parameter_map)
```
josh146: This will close #7 I just realised 🙂

pennylane_forest/qpu.py
```diff
@@ -82,6 +82,14 @@ def __init__(self, device, *, shots=1024, active_reset=True, load_qc=True, reado
...
+            self._parameter_map = {}
+            self._parameter_reference_map = {}
+
+        timeout = kwargs.pop("timeout", None)
```
josh146: Same comment here as the QVM: both timeout and parametric_compilation should be added to the docstring

requirements.txt
```diff
@@ -1,3 +1,4 @@
 pyquil>=2.16
-pennylane>=0.6
+git+https://github.com/XanaduAI/pennylane.git#egg=pennylane
```
josh146: Let's not forget to revert this before release!

tests/test_qpu.py
```diff
@@ -2,18 +2,21 @@
...
+from conftest import BaseTest, QVM_SHOTS
 
+from flaky import flaky
 
 log = logging.getLogger(__name__)
 
-VALID_QPU_LATTICES = [qc for qc in pyquil.list_quantum_computers() if "qvm" not in qc]
-
+pattern = 'Aspen-.-[1-5]Q-.'
+VALID_QPU_LATTICES = [qc for qc in pyquil.list_quantum_computers() if "qvm" not in qc and re.match(pattern, qc)]
```
josh146: Good catch. For context: previously, VALID_QPU_LATTICES would return reasonably small sized lattices for the test suite to run on. However, with Aspen-7 now available, a lot of these lattices were significantly larger, causing the test suite to hang/run out of memory 🙂

tests/test_qpu.py
```diff
@@ -56,6 +59,7 @@ class TestQPUBasic(BaseTest):
     # pylint: disable=protected-access
 
     def test_no_readout_correction(self):
+        """Test the QPU plugin with no readout correction"""
```
josh146: Thanks for catching these missing docstrings!

tests/test_qpu.py
```diff
@@ -151,10 +156,16 @@ def circuit_Zmi():
         assert np.allclose(results[:3], 1.0, atol=2e-2)
         assert np.allclose(results[3:], -1.0, atol=2e-2)
 
+    @flaky(max_runs=10, min_passes=1)
```
josh146: Since flaky should only be used when absolutely necessary, perhaps it's worth justifying in the docstring. Is it just because the number of shots here are small? 1 pass out of 10 runs doesn't seem like a high enough pass rate 😨 (I would imagine something like `max_runs=5`, `min_passes=3` being a lot more reasonable)

josh146: This comment applies to all following uses of flaky

josh146: Just realised this comment refers to a change that was made in #38, not here.

tests/test_qpu.py
```diff
@@ -163,3 +174,82 @@ def circuit():
...
+        dev_qpu = qml.device('forest.qpu', device=device, load_qc=False, readout_error=[0.9, 0.75],
+                            symmetrize_readout="exhaustive", calibrate_readout="plus-eig", shots=QVM_SHOTS)
+
+        @qml.qnode(dev_qpu)
+        def circuit():
+            qml.RY(np.pi/2, wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0) @ qml.Identity(1))
+
+        assert np.allclose(circuit(), 0.0, atol=2e-2)
```
josh146: A better test might be a circuit that takes a parameter. For example, 
```python
@qml.qnode(dev_qpu)
def circuit(x):
    qml.RY(x, wires=[0])
    qml.Hadmard(wires=1)
    qml.CNOT(wires=[0, 1])
    return qml.expval(qml.PauliZ(0) @ qml.Identity(1))

assert np.allclose(circuit(a), np.cos(a), atol=2e-2)
```
This could then be parametrized for different values of `a`

tests/test_qpu.py
```diff
@@ -163,3 +174,82 @@ def circuit():
...
+        dev_qpu = qml.device('forest.qpu', device=device, load_qc=False, readout_error=[0.9, 0.75],
+                            symmetrize_readout="exhaustive", calibrate_readout="plus-eig", shots=QVM_SHOTS)
+
+        @qml.qnode(dev_qpu)
+        def circuit():
+            qml.Hadamard(0)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+        assert np.allclose(circuit(), 1.0, atol=2e-2)
```
josh146: Similar comment to above:
```python
@qml.qnode(dev_qpu)
def circuit(x):
    qml.RY(x, wires=[0])
    qml.Hadmard(wires=1)
    qml.CNOT(wires=[0, 1])
    return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))

assert np.allclose(circuit(a), np.cos(a), atol=2e-2)
```

tests/test_qpu.py
```diff
@@ -163,3 +174,82 @@ def circuit():
...
+
+    @flaky(max_runs=10, min_passes=1)
+    def test_2q_gate_pauliz_pauliz_tensor_parametric_compilation_off(self):
+        """Test that the PauliZ tensor PauliZ observable works correctly, when parametric compilation
+        was turned off.
```
josh146:
```suggestion
        is turned off.
```

tests/test_qpu.py
josh146: It might also be good to call the assert one more time? Just to check that repeated calling of the QNode works correctly?

tests/test_qvm.py
```diff
@@ -386,6 +501,203 @@ def test_sample_values_hermitian_multi_qubit(self, qvm, tol):
...
+    @pytest.mark.parametrize("device", ["2q-qvm", np.random.choice(VALID_QPU_LATTICES)])
+    def test_timeout_default(self, shots, device):
+        """Test that the timeout attrbiute for the QuantumComputer stored by the QVMDevice
+        is set correctly when passing a value as keyword argument"""
+        dev = plf.QVMDevice(device=device, shots=shots)
+        qc = pyquil.get_qc(device, as_qvm=True)
+
+        # Check that the timeouts are equal (it has not been changed as a side effect of
+        # instantiation
+        assert dev.qc.compiler.client.timeout == qc.compiler.client.timeout
```
josh146: nice tests!

tests/test_qvm.py
```diff
@@ -386,6 +501,203 @@ def test_sample_values_hermitian_multi_qubit(self, qvm, tol):
...
+        circuit_graph = CircuitGraph([
+                                       qml.RX(theta, wires=[0]),
+                                       qml.RX(phi, wires=[1]),
+                                       qml.CNOT(wires=[0, 1])
+                                       ],
+                                         [
+                                        O1,
+                                        O2
+                                        ]
+                                    )
```
josh146: This looks a bit funny - don't forget to `black -l 100` before this is merged!

tests/test_qvm.py
```diff
@@ -386,6 +501,203 @@ def test_sample_values_hermitian_multi_qubit(self, qvm, tol):
...
+        # Calling generate_samples with unchanged hash
+        for i in range(6):
+            with monkeypatch.context() as m:
+                m.setattr(QuantumComputer, "compile", lambda self, prog: call_history.append(prog))
+                m.setattr(QuantumComputer, "run", lambda self, **kwargs: None)
+                dev.generate_samples()
+
+            assert dev.circuit_hash in dev._lookup_table
+            assert len(dev._lookup_table.items()) == 1
+            assert len(call_history) == 1
```
josh146: Awesome :)

tests/test_qvm.py
```diff
@@ -475,3 +791,209 @@ def circuit(x, y, z):
...
+
+        qnodes = qml.map(circuit, obs_list, dev)
+
+        for idx, stmt in enumerate(statements):
+            qnodes[idx]([], statement=stmt)
+            assert dev.circuit_hash in dev._lookup_table
+
+        # Using that True evaluates to 1
+        number_of_true = sum(statements)
+        length = 1 if (number_of_true == 6 or number_of_true == 0) else 2
```
josh146: I'm not quite sure I follow the logic in this line 🤔

tests/test_qvm.py
```diff
@@ -475,3 +791,209 @@ def circuit(x, y, z):
...
+        obs = [qml.PauliZ(0) @ qml.PauliZ(1)]
+        obs_list = obs * 6
+
+        qnodes = qml.map(circuit, obs_list, dev)
+
+        for idx, qnode in enumerate(qnodes):
+            qnode([], rounds=idx)
+            assert dev.circuit_hash in dev._lookup_table
+
+        assert len(dev._lookup_table.items()) == len(qnodes)
```
josh146: really nice!

tests/test_qvm.py
```diff
@@ -475,3 +791,209 @@ def circuit(x, y, z):
...
+        results = qnodes(params)
+
+        assert len(call_history) == 0
+        assert dev.circuit_hash in dev._lookup_table
+        assert len(dev._lookup_table.items()) == 1
+
+    @flaky(max_runs=5, min_passes=1)
+    @pytest.mark.parametrize("device", ["2q-qvm", np.random.choice(VALID_QPU_LATTICES)])
+    def test_compiled_program_was_correct_compared_with_default_qubit(self, qvm, device, tol):
+        """Test that QVM device stores the compiled program correctly by comparing it with default.qubit.
```
josh146:
```suggestion
        """Test that QVM device executes the compiled program correctly by comparing numerical output with default.qubit.
```

tests/test_qvm.py
```diff
@@ -475,3 +791,209 @@ def circuit(x, y, z):
...
+
+        assert np.allclose(results, results2, atol=1e-02, rtol=0)
+        assert dev.circuit_hash in dev._lookup_table
+        assert len(dev._lookup_table.items()) == 1
+
+    @flaky(max_runs=10, min_passes=1)
+    @pytest.mark.parametrize("device", ["2q-qvm", np.random.choice(VALID_QPU_LATTICES)])
+    def test_2q_gate_pauliz_pauliz_tensor_parametric_compilation_off(self, device, qvm, compiler):
+        """Test that the PauliZ tensor PauliZ observable works correctly, when parametric compilation
+        was turned off.
```
josh146:
```suggestion
        is turned off.
```

pennylane_forest/device.py
antalszava: The parameter map stores the numeric value of the symbolic parameter, whereas the parameter reference map stores the PyQuil reference that has been created for it. I'll add docstrings explaining these differences

pennylane_forest/qvm.py
antalszava: Changed here and in ``device.py``!

pennylane_forest/qvm.py
antalszava: Yes, this flow of logic is good! Added some comments, let me know if it's still unclear

pennylane_forest/qvm.py
antalszava: Such that ``self._parameter_map`` can be passed later as required by the ``QuantumComputer.run`` method:
``self.qc.run(executable=compiled_program, memory_map=self._parameter_map)``

Here the values should be a list of memory references as here:
http://docs.rigetti.com/en/stable/migration3-declare.html#Using-memory_map

pennylane_forest/qvm.py
antalszava: Sure, sounds really good!

tests/test_qpu.py
antalszava: Still adjusted them in there, thanks!

tests/test_qvm.py
antalszava: Will change it. The main point here is that ``True`` evaluates to 1 whereas ``False`` evaluates as 0.

antalszava: This checks if all elements in the list were either ``True`` or ``False`` because in such a case we have compiled only one program.

tests/test_qpu.py
antalszava: Added a parametric test here with two assertions each having one call on ``circuit``

pennylane_forest/device.py
```diff
@@ -244,21 +243,36 @@ def apply(self, operations, **kwargs):
...
+        """Apply the circuit rotations.
+
+        This method serves as an auxiliary method to :meth:`~.ForestDevice.apply`.
+
+        Args:
+            rotations (List[pennylane.Operation]): operations that rotate into the
+                measurement basis
+
+        Returns:
+            list: wires as integers corresponding to the wiring if applicable
```
josh146: doesn't this method return a pyQuil program?

pennylane_forest/qpu.py
```diff
@@ -68,6 +68,9 @@ class QPUDevice(QVMDevice):
         compiler_url (str): the compiler server URL. Can also be set by the environment
             variable ``COMPILER_URL``, or in the ``~/.forest_config`` configuration file.
             Default value is ``"http://127.0.0.1:6000"``.
+        timeout (int): Number of seconds to wait for a response from the client.
+        parametric_compilation (bool): a boolean value of whether or not to use parametric
```
josh146:
```suggestion
        parametric_compilation (bool): whether or not to use parametric compilation.
```

pennylane_forest/qpu.py
```diff
@@ -68,6 +68,9 @@ class QPUDevice(QVMDevice):
...
+            compilation. It is True by default.
```
josh146: No need to specify a default :)
```suggestion
```

pennylane_forest/qpu.py
```diff
@@ -83,6 +86,28 @@ def __init__(self, device, *, shots=1024, active_reset=True, load_qc=True, reado
...
+
+        self.parametric_compilation = kwargs.get("parametric_compilation", True)
+
+        if self.parametric_compilation:
+            self._compiled_program_dict = {}
+            """dict[int, pyquil.ExecutableDesignator]: stores circuit hashes associated
+                with the corresponding compiled programs."""
+
+            self._parameter_map = {}
+            """dict: stores the string of symbolic parameters associated with
```
josh146:
```suggestion
            """dict[str, float]: stores the string of symbolic parameters associated with
```

pennylane_forest/qpu.py
```diff
@@ -83,6 +86,28 @@ def __init__(self, device, *, shots=1024, active_reset=True, load_qc=True, reado
...
+                their numeric values. This map will be used to bind parameters in a parametric
+                program using PyQuil."""
+
+            self._parameter_reference_map = {}
+            """dict: stores the string of symbolic parameters associated with
```
josh146: What are the types of the dictionary keys/values here?

pennylane_forest/qpu.py
```diff
@@ -83,6 +86,28 @@ def __init__(self, device, *, shots=1024, active_reset=True, load_qc=True, reado
...
+        self._compiled_program = None
+        """Union[None, pyquil.ExecutableDesignator]: the latest compiled program. If parametric
+        compilation is turned on, this will be a parametric program."""
+
+        self.parametric_compilation = kwargs.get("parametric_compilation", True)
+
+        if self.parametric_compilation:
+            self._compiled_program_dict = {}
+            """dict[int, pyquil.ExecutableDesignator]: stores circuit hashes associated
+                with the corresponding compiled programs."""
```
josh146: Since the key is the hash, shouldn't it be a `str`, not an `int`?

pennylane_forest/qvm.py
```diff
@@ -62,6 +65,9 @@ class QVMDevice(ForestDevice):
         compiler_url (str): the compiler server URL. Can also be set by the environment
             variable ``COMPILER_URL``, or in the ``~/.forest_config`` configuration file.
             Default value is ``"http://127.0.0.1:6000"``.
+        timeout (int): Number of seconds to wait for a response from the client.
+        parametric_compilation (bool): a boolean value of whether or not to use parametric
+            compilation. It is True by default.
```
josh146: Same comment here as in QPU :)

pennylane_forest/qvm.py
```diff
@@ -75,6 +81,27 @@ def __init__(self, device, *, shots=1024, noisy=False, **kwargs):
...
+                with the corresponding compiled programs."""
+
+            self._parameter_map = {}
+            """dict[string, complex]: stores the string of symbolic parameters associated with
+                their numeric values. This map will be used to bind parameters in a parametric
+                program using PyQuil."""
+
+            self._parameter_reference_map = {}
+            """dict[string, pyquil.quilatom.MemoryReference]: stores the string of symbolic
+                parameters associated with their PyQuil memory references."""
```
josh146: Same comments on the docstrings as in QPU :)

pennylane_forest/qvm.py
```diff
@@ -133,9 +165,95 @@ def apply(self, operations, **kwargs):
...
+
+        if self.circuit_hash not in self._compiled_program_dict:
+            # Compiling this specific program for the first time
+            # Store the compiled program with the corresponding hash
+            self._compiled_program_dict[self.circuit_hash] = self.qc.compile(self.prog)
+
+        # The program has been compiled, store as the latest compiled program
+        self._compiled_program = self._compiled_program_dict[self.circuit_hash]
+        samples = self.qc.run(executable=self._compiled_program, memory_map=self._parameter_map)
+        return samples
```
josh146: Much easier to follow with the comments!

pennylane_forest/qvm.py
```diff
@@ -133,9 +165,95 @@ def apply(self, operations, **kwargs):
...
+
+        If parametric compilation is turned on, this will be a parametric program.
+
+        The pyquil.ExecutableDesignator.program attribute stores the pyquil.Program
+        instance. If no program was compiled yet, this property returns None.
+
+        Returns:
+            Union[None, pyquil.ExecutableDesignator]: the latest compiled program
+        """
+        return self._compiled_program
```
josh146: Nice having this property

tests/test_qpu.py
```diff
@@ -1,200 +1,265 @@
...
+        @qml.qnode(dev_qpu)
+        def circuit(x):
+            qml.RY(x, wires=[0])
+            qml.Hadamard(wires=1)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0) @ qml.Identity(1))
+
+        assert np.allclose(circuit(a), np.cos(a), atol=2e-2)
+        # Check that repeated calling of the QNode works correctly
+        assert np.allclose(circuit(a), np.cos(a), atol=2e-2)
```
josh146: 🏆

tests/test_qpu.py
```diff
@@ -1,200 +1,265 @@
...
+            return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+        analytic_value = np.cos(a/2)**2 * np.cos(b/2) ** 2\
+                        + np.cos(b/2) ** 2 * np.sin(a/2) ** 2\
+                        - np.cos(a/2) ** 2 * np.sin(b/2) ** 2\
+                        - np.sin(a/2) ** 2 * np.sin(b/2) ** 2
+
+        assert np.allclose(circuit(a, b), analytic_value, atol=2e-2)
+        # Check that repeated calling of the QNode works correctly
+        assert np.allclose(circuit(a, b), analytic_value, atol=2e-2)
```
josh146: Nice test!

tests/test_qpu.py
```diff
@@ -1,200 +1,265 @@
...
+        assert np.allclose(circuit(a), np.cos(a), atol=2e-2)
+
+    @flaky(max_runs=10, min_passes=1)
+    @pytest.mark.parametrize("a", np.linspace(-np.pi/2, 0, 3))
+    @pytest.mark.parametrize("b", np.linspace(0, np.pi/2, 3))
+    def test_2q_gate_pauliz_pauliz_tensor_parametric_compilation_off(self, a, b):
+        """Test that the PauliZ tensor PauliZ observable works correctly, when parametric compilation
+        is turned off.
+
+        As the results coming from the qvm are stochastic, a constraint of  3 out of 5 runs was added.
```
josh146: This doesn't match the flaky decorator (which says 1/10 tests)

pennylane_forest/device.py
antalszava: Oh you're right, thanks for that!

pennylane_forest/qpu.py
antalszava: Thanks for getting that! Forgot to update this occurrence, whereas the ``QVMDevice`` counterpart has the docstring straight (one extra reason for having these defined at only one place)

antalszava: (Getting at the ``string`` -> ``str`` part as well from ``QVM``)

pennylane_forest/qpu.py
antalszava: The hash is an ``int`` that was created by calling on ``hash`` with the serialized ``circuit_graph`` as the argument

josh146: Ah, of course!

pennylane_forest/qvm.py
antalszava: Thanks to a keen user for that!

tests/test_qpu.py
antalszava: Thanks for getting that!

kilimanjaro:
Thanks for the detailed responses. Those cleared things up for me!

PR #41, opened 2020-02-04
Testing single observable Tensor operator estimation, mutual exclusion with parametric compilation

antalszava:
**Context:**
* Previously errors arose when using single observable `Tensor` instances in qpu.expval
* Currently, operator estimation uses the older version of PyQuil API which supports non-parametric programs. This causes operator estimation will fail as it would like to use a `Program` with undefined PyQuil memory references.

* Adding more tests as related to https://github.com/XanaduAI/pennylane/pull/498 PR from PennyLane core
* Making parametric compilation and operator estimation mutually exclusive

**Benefits:**
* `Tensor` instances made up of a single observable can now be passed to the `QPUDevice.expval` method
* No conflict between parametric compilation and operator estimation

**Possible Drawbacks:**
* Operator estimation is disabled at times for now

**Related GitHub Issues:**
N/A

### DIFF
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -153,16 +153,9 @@ class QPUDevice(QVMDevice):
 
     def expval(self, observable):
         wires = observable.wires
+
         # Single-qubit observable
         if len(wires) == 1:
-
-            # In case the observable is a Tensor, then
-            # it is made up of exactly one observable
-            # so we can extract it to perform operator
-            # estimation
-            if isinstance(observable, Tensor):
-                observable = observable.obs[0]
-                wires = observable.wires
 
             # identify Experiment Settings for each of the possible single-qubit observables
             wire = wires[0]
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -166,8 +166,9 @@ class QPUDevice(QVMDevice):
     def expval(self, observable):
         wires = observable.wires
 
-        # Single-qubit observable
         if len(wires) == 1 and not self.parametric_compilation:
+            # Single-qubit observable when parametric compilation is turned off
+
             # identify Experiment Settings for each of the possible single-qubit observables
             wire = wires[0]
             qubit = self.wiring[wire]
@@ -181,8 +182,10 @@ class QPUDevice(QVMDevice):
                     ExperimentSetting(TensorProductState(), float(np.sqrt(1 / 2)) * sZ(qubit)),
                 ],
             }
-            # expectation values for single-qubit observables
+
             if observable.name in ["PauliX", "PauliY", "PauliZ", "Identity", "Hadamard"]:
+                # expectation values for single-qubit observables
+
                 prep_prog = Program()
                 for instr in self.program.instructions:
                     if isinstance(instr, Gate):
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -158,7 +158,7 @@ class QPUDevice(QVMDevice):
 
             # In case the observable is a Tensor, then
             # it is made up of exactly one observable
-            # so we can extract that to perform operator
+            # so we can extract it to perform operator
             # estimation
             if isinstance(observable, Tensor):
                 observable = observable.obs[0]
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -24,9 +24,13 @@ import warnings
 
 import numpy as np
 from pyquil import get_qc
-from pyquil.operator_estimation import (Experiment, ExperimentSetting,
-                                        TensorProductState, group_experiments,
-                                        measure_observables)
+from pyquil.operator_estimation import (
+    Experiment,
+    ExperimentSetting,
+    TensorProductState,
+    group_experiments,
+    measure_observables,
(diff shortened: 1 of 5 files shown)


tests/test_qpu.py
```diff
@@ -55,6 +56,39 @@ def test_qpu_args(self):
...
+
+        @qml.qnode(dev_1)
+        def circuit_obs(param):
+            template(param)
+            return qml.expval(obs)
+
+        res = circuit_tensor(p)
+        exp = circuit_obs(p)
+
+        assert np.allclose(res, exp, atol=2e-2)
```
josh146: Nice test!

josh146:
@kilimanjaro, it looks like Semaphore is failing with `No QVM server running at http://localhost:5000`, which is quite strange. @antalszava, perhaps if you retrigger the build?

pennylane_forest/qpu.py
```diff
@@ -153,6 +155,15 @@ def expval(self, observable):
...
         # Single-qubit observable
         if len(wires) == 1:
+
+            # In case the observable is a Tensor, then
+            # it is made up of exactly one observable
+            # so we can extract it to perform operator
+            # estimation
+            if isinstance(observable, Tensor):
+                observable = observable.obs[0]
+                wires = observable.wires
```
msohaibalam: Is it intentional that this definition of `wires` gives a list of list, while the one appearing directly above the `if len(wires)==1` condition gives just a list? I'm wondering both (a) whether the list of list structure messes with the intended manipulations below, and (b) if we can just continue to use the already defined `wires`.

msohaibalam:
Looks good to me!

pennylane_forest/qpu.py
antalszava: This is an excellent question! We have decided to address this in PennyLane core and merged the following PR:
https://github.com/XanaduAI/pennylane/pull/498

This PR essentially entails the type of logic contained in the commented lines and treats another edge case:
a) if `Tensor` contains a single observable, then this single observable is returned
b) it also removes any and all `Identity` observables in a `Tensor`. The reason for this is that a `Tensor` could contain *exactly one* non-Identity observable, and yet behave as a `Tensor` containing multiple such observables. This would then not apply operator estimation to these observables

Breaking up a `Tensor` instance along these lines will now happen inside of PennyLane as the corresponding part of the `QNode` was also modified. This essentially means, that by the time the program execution hits this part, there will be no need for such checks here anymore.

pennylane_forest/qpu.py
```diff
@@ -106,6 +98,17 @@ def __init__(
         """Union[None, pyquil.ExecutableDesignator]: the latest compiled program. If parametric
         compilation is turned on, this will be a parametric program."""
 
+
+        if "parametric_compilation" in kwargs and kwargs["parametric_compilation"]:
```
josh146: This can probably be simplified to
```suggestion
        if kwargs.get("parametric_compilation", False):
```

josh146: (Is this False by default?)

pennylane_forest/qpu.py
```diff
@@ -106,6 +98,17 @@ def __init__(
...
+            # Raise a warning if parametric compilation was explicitly turned on by the user
+            # about turning the operator estimation off
+
+            # TODO: Remove the warning and togglig once a migration to the new operator estimation
```
josh146:
```suggestion
            # TODO: Remove the warning and toggling once a migration to the new operator estimation
```

tests/test_pyqvm.py
```diff
@@ -365,7 +365,7 @@ def circuit():
     @pytest.mark.parametrize("device", ["2q-pyqvm"])
     def test_one_qubit_wavefunction_circuit(self, device, shots):
         """Test that the wavefunction plugin provides correct result for simple circuit."""
-        shots = 100000
+        shots = 10000
```
josh146: Has this just been decreased to reduce the test suite time?

tests/test_pyqvm.py
```diff
@@ -380,4 +380,4 @@ def circuit(x, y, z):
             qml.Rot(x, y, z, wires=0)
             return qml.expval(qml.PauliZ(0))
 
-        self.assertAlmostEqual(circuit(a, b, c), np.cos(a) * np.sin(b), delta=3 / np.sqrt(shots))
+        self.assertAlmostEqual(circuit(a, b, c), np.cos(a) * np.sin(b), delta=5 / np.sqrt(shots))
```
josh146: Rather than increase the std deviation, should we make this test flaky?

pennylane_forest/qpu.py
antalszava: It's not included in the signature, later defaults to `True`. Here we would like to check for the case when the user has explicitly turned on parametric compilation

tests/test_pyqvm.py
antalszava: Yes. In the previous run, for some reason, the testing was killed at one of the PyQvm tests

tests/test_pyqvm.py
antalszava: I have the worry that it would take even more time, currently these tests from `pyqvm.py` take a while to run

PR #42, opened 2020-02-07
Adding support for multi-qubit observable estimation

msohaibalam:
Extends `qpu.expval` to handle multi-qubit observable estimation.

Handles the scenarios where (a) multi-qubit observable come in the `Tensor` format, and (b) pre-rotations are appropriately applied to all qubits so that we only need to measure the PauliZ operator.

Cleans up and consolidates the logic to remove code repetition between single and multi-qubit observable estimation.

A few things to note:
- While the wiring is no longer needed in writing the preparation program, it is still needed when defining one's own readout errors on a qpu-as-a-qvm.
- The same readout errors are defined on all the qubits via `readout_error=[p00, p11]` for simplicity. If we need the ability to define different readout errors on distinct qubits, we might choose to promote `readout_error` to be a list of lists in the multi-qubit case and handle that scenario separately.


pennylane_forest/qpu.py
```diff
@@ -205,4 +216,51 @@ def expval(self, observable):
...
                 return w[0] * p0 + w[1] * p1
 
+        # Multi-qubit observable
+        if len(wires) > 1:
+            # In this case, the observable must be a Tensor product
+            assert isinstance(observable, Tensor), "Multi-qubit observable is not in Tensor format"
+
+            # All observables are rotated to be measured in the Z-basis, so we just need to
+            # check which wires exist in the observable, map them to physical qubits, and measure
+            # the product of PauliZ operators on those qubits
```
josh146: Nice!

pennylane_forest/qpu.py
```diff
@@ -205,4 +216,51 @@ def expval(self, observable):
...
+            grouped_tomo_expt = group_experiments(tomo_expt)
+            meas_obs = list(
+                measure_observables(
+                    self.qc,
+                    grouped_tomo_expt,
+                    active_reset=self.active_reset,
+                    symmetrize_readout=self.symmetrize_readout,
+                    calibrate_readout=self.calibrate_readout,
+                )
+            )
```
josh146: Something that @antalszava pointed out is that `measure_observables` will implicitly execute the circuit on the QPU. However, `QVMDevice.generate_samples()` has already done one execution, making it redundant. This could be solved by simply overwriting `generate_samples()` here to take into account operator estimation, _or_ perhaps by our eventual goal of merging QVM and QPU :laughing:

tests/test_qpu.py
```diff
@@ -55,6 +56,36 @@ def test_qpu_args(self):
...
+
+        @qml.qnode(dev_1)
+        def circuit_obs(param):
+            template(param)
+            return qml.expval(obs)
+
+        res = circuit_tensor(p)
+        exp = circuit_obs(p)
+
+        assert np.allclose(res, exp, atol=2e-2)
```
josh146: Awesome!

tests/test_qvm.py
```diff
@@ -964,17 +969,17 @@ def test_compiled_program_was_correct_compared_with_default_qubit(self, qvm, dev
 
         results2 = qnodes2(params)
 
-        assert np.allclose(results, results2, atol=1e-02, rtol=0)
+        assert np.allclose(results, results2, atol=2e-02, rtol=0)
         assert dev.circuit_hash in dev._compiled_program_dict
         assert len(dev._compiled_program_dict.items()) == 1
 
-    @flaky(max_runs=10, min_passes=1)
+    @flaky(max_runs=5, min_passes=3)
```
josh146: While not best practice for unittests, we've found flaky really nice for stochastic quantum integration tests :laughing:

pennylane_forest/qpu.py
```diff
@@ -153,6 +155,15 @@ def expval(self, observable):
...
         # Single-qubit observable
         if len(wires) == 1:
+
+            # In case the observable is a Tensor, then
+            # it is made up of exactly one observable
+            # so we can extract it to perform operator
+            # estimation
+            if isinstance(observable, Tensor):
+                observable = observable.obs[0]
+                wires = observable.wires
```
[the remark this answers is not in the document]
msohaibalam: Oh yeah, this is just code from @antalszava's PR which I decided to build off instead of master. I'll merge in the new changes.

pennylane_forest/qpu.py
```diff
@@ -205,4 +216,51 @@ def expval(self, observable):
                 w = self._eigs[Hkey]["eigval"]
                 return w[0] * p0 + w[1] * p1
 
+        # Multi-qubit observable
+        if len(wires) > 1:
+            # In this case, the observable must be a Tensor product
+            assert isinstance(observable, Tensor), "Multi-qubit observable is not in Tensor format"
```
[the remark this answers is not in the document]
msohaibalam: Got it! It sounds like then that the check for `observable.name == "Hermitian"` should handle both the single-qubit and multi-qubit observable cases then. I'll make the change.

tests/test_qpu.py
msohaibalam: Borrowed code from @antalszava's PR :)

tests/test_qvm.py
msohaibalam: Seems like a nice tool!

pennylane_forest/qpu.py
msohaibalam: Not sure I follow this point. Is the idea `QVMDevice.generate_samples()` being called before/while this particular call to `measure_observables`?  (and if that's the case, it's not too clear to me where exactly this might be happening)

antalszava: Yes, one missing piece here is that we have moved to a new API in PennyLane and this affects the plugins as well.

Following up on the previous https://github.com/rigetti/pennylane-forest/pull/38, now ForestDevice inherits from QubitDevice. The PennyLane type of flow of logic involves evaluating a quantum circuit by calling on the execute method of the device (such is QPUDevice).

Each instance of the ForestDevice (now QPUDevice in our case) will currently inherit the same execute method [as seen here](https://github.com/XanaduAI/pennylane/blob/a1f22b10ee968ade10488c6c42a7b153553cf628/pennylane/_qubit_device.py#L123) (which is not redefined in any of the subclasses).

As per the logic in `QubitDevice.execute`, currently [`generate_samples` is called](https://github.com/XanaduAI/pennylane/blob/8cccf8681ccfc81f46a771092fcb4442e0319512/pennylane/_qubit_device.py#L158), as the `QPUDevice` does not support analytic evaluation (it being a hardware). After this, [statistics are extracted](https://github.com/XanaduAI/pennylane/blob/8cccf8681ccfc81f46a771092fcb4442e0319512/pennylane/_qubit_device.py#L161) and therefore eventually we [call on `QPUDevice.expval`](https://github.com/XanaduAI/pennylane/blob/a1f22b10ee968ade10488c6c42a7b153553cf628/pennylane/_qubit_device.py#L249) which usually corresponds to `QubitDevice.expval`.

Now in our case, `QPUDevice` inherited the `QVMDevice.generate_samples` method, which makes the call on the `QuantumComputer.run` method. It will in the `expval` method, however, call on `measure_observables` as well.

msohaibalam: Thanks for the detailed explanation @antalszava! It seems to me this won't be the last time `generate_samples` and `measure_observables` will conflict ;) , mostly because they're both trying to generate samples!

My gut feeling is also that even if we do merge the QVM and QPU classes, we'll still be faced with the choice of either over-riding `generate_samples` to use `measure_observables`, or port over all the logic of error mitigation over to `generate_samples` (and/or its helper functions). I'm not sure how I can over-ride it here though, would appreciate input on that!

msohaibalam:
Thanks for the feedback, @josh146 and @antalszava! I'm done making changes and updates, so please give this another look when you can.

Re: operator_estimation not handling parametric compilation, yes that is unfortunately a feature currently lacking, and one of the aims of the new API (once it's ready) is to include precisely this functionality.

pennylane_forest/qpu.py
```diff
@@ -106,6 +102,17 @@ def __init__(
...
+
+        if kwargs.get("parametric_compilation", False):
+            # Raise a warning if parametric compilation was explicitly turned on by the user
+            # about turning the operator estimation off
+
+            # TODO: Remove the warning and toggling once a migration to the new operator estimation
+            # API has been executed. This new API provides compatibility between parametric
+            # compilation and operator estimation.
+            warnings.warn("Parametric compilation is currently not supported with operator"
+                          "estimation. Operator estimation is being turned off.")
```
josh146: How come this is showing up as a change, has the location been moved?

pennylane_forest/qpu.py
```diff
@@ -20,28 +20,24 @@
...
 
 import numpy as np
-
 from pyquil import get_qc
-from pyquil.api._quantum_computer import _get_qvm_with_topology
-from pyquil.gates import MEASURE, RESET
-from pyquil.quil import Pragma, Program
+from pyquil.operator_estimation import (Experiment, ExperimentSetting,
+                                        TensorProductState, group_experiments,
+                                        measure_observables)
```
josh146: Nice to have the updated import locations!

pennylane_forest/qpu.py
```diff
@@ -151,58 +158,75 @@ def __init__(
...
-                            str_instr += f" {int(w)}"
-                        prep_prog += Program(str_instr)
-
-                if self.readout_error is not None:
+
+        if observable.name == "Hermitian":
+            # <H> = \sum_i w_i p_i
+            Hkey = tuple(par[0].flatten().tolist())
+            w = self._eigs[Hkey]["eigval"]
+            return w[0] * p0 + w[1] * p1
```
josh146: This might not be needed, as the call to `super().expval(observable)` already takes into account this case 🙂

pennylane_forest/qpu.py
```diff
@@ -151,58 +158,75 @@ def __init__(
...
+
+        # `measure_observables` called only when parametric compilation is turned off
+        elif not self.parametric_compilation:
```
josh146: In which case, this can simply become `if not self.parametric_compilation:`

pennylane_forest/qpu.py
```diff
@@ -151,58 +158,75 @@ def __init__(
...
+            # Multi-qubit observable
+            elif len(wires) > 1 and isinstance(observable, Tensor) and not self.parametric_compilation:
+
+                # All observables are rotated to be measured in the Z-basis, so we just need to
+                # check which wires exist in the observable, map them to physical qubits, and measure
+                # the product of PauliZ operators on those qubits
+                pauli_obs = sI()
+                for wire in observable.wires:
+                    qubit = wire[0]
+                    pauli_obs *= sZ(self.wiring[qubit])
```
josh146: Awesome!

pennylane_forest/qpu.py
josh146: > I'm not sure how I can over-ride it here though, would appreciate input on that!

I think it's a good question, and not entirely trivial --- I say we merge this as is, and make a new issue to determine how to avoid this conflict!

msohaibalam: That sounds good to me. I've created a new issue here: https://github.com/rigetti/pennylane-forest/issues/45.

pennylane_forest/qpu.py
msohaibalam: I changed this to have the same format as on the master branch.

pennylane_forest/qpu.py
msohaibalam: Great! Removing this code block.

pennylane_forest/qpu.py
msohaibalam: Updated.

PR #44, opened 2020-02-09
Bump version number and update requirements

josh146:
* Updates the version number to v0.8, to bring it inline with the latest PennyLane release

* Updates the setup.py and requirements.txt, to specify that the plugin now only works with PennyLane v0.8 and above.

### DIFF
--- a/pennylane_forest/device.py
+++ b/pennylane_forest/device.py
@@ -178,9 +178,9 @@ class ForestDevice(QubitDevice):
             variable ``QUILC_URL``, or in the ``~/.forest_config`` configuration file.
             Default value is ``"http://127.0.0.1:6000"``.
     """
-    pennylane_requires = ">=0.6"
+    pennylane_requires = ">=0.8"
     version = __version__
-    author = "Josh Izaac"
+    author = "Rigetti Computing Inc."
 
     _operation_map = pyquil_operation_map
     _capabilities = {"model": "qubit", "tensor_observables": True}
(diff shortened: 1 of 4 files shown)


pennylane_forest/device.py
```diff
@@ -178,9 +178,9 @@ class ForestDevice(QubitDevice):
             variable ``QUILC_URL``, or in the ``~/.forest_config`` configuration file.
             Default value is ``"http://127.0.0.1:6000"``.
     """
-    pennylane_requires = ">=0.6"
+    pennylane_requires = ">=0.8"
     version = __version__
-    author = "Josh Izaac"
+    author = "Rigetti Computing Inc."
```
kilimanjaro: I don't know exactly who all worked on this, but I don't think it's unfair to give @josh146 some of the credit!

josh146: This is part of the information that shows up when you query the device information in PennyLane. I thought best to sync it with the author field in `setup.py` :)

I think this field hasn't changed since I first prototyped this plugin, so definitely worth updating!

msohaibalam: I think it'd definitely be unfair not to give @josh146 credit! It means we should probably update `setup.py` as well. I would've been fine just to stick with the original author (i.e. Josh), but if the view is to include everyone who's authored the library since, maybe something like "Rigetti, Xanadu" or even "Rigetti, Xanadu and friends" (since we now have contributions from others as well)?

PR #49, opened 2020-04-06
Correcting tests

antalszava:
**Context:**
A number of tests fail while running the test suite due to the following reasons:
* The `"Aspen-4-4Q-E"` lattice, that was specifically used by certain tests, is no longer a valid `QuantumComputer` lattice;
* The default number of shots has been set to `1000` for `QubitDevice` in PL core;
* The `Variable.free_param_values` has been renamed in PL core.
 
The following parts of the test suite were updated:
* Changed to picking a valid four qubit lattice at runtime instead of hardcoding it in QVM tests;
* Updated the checks for the number of shots;
* Updated the name of the Variable class attribute for argument values coming from PL core.

**Benefits:**
The test suite runs well again.

**Possible drawbacks:**
The name of the Variable class attribute was used and this might potentially break again, should it be subject to another change in PL core.

### DIFF
--- a/pennylane_forest/device.py
+++ b/pennylane_forest/device.py
@@ -359,3 +359,28 @@ class ForestDevice(QubitDevice):
 
         return np.reshape(state_multi_index, 2 ** self.num_wires)
 
+    def analytic_probability(self, wires=None):
+        """Return the (marginal) probability of each computational basis
+        state from the last run of the device.
+
+        If no wires are specified, then all the basis states representable by
+        the device are considered and no marginalization takes place.
+
+        .. warning:: This method will have to be redefined for hardware devices, since it uses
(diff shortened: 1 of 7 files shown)


josh146:
Thanks @antalszava! It looks like some of the tests are failing due to 'flakiness'; did you want to push another commit to rerun semaphore?

pennylane_forest/device.py
```diff
@@ -359,7 +359,7 @@ def mat_vec_product(self, mat, vec, wires):
 
         return np.reshape(state_multi_index, 2 ** self.num_wires)
 
-    def probability(self, wires=None):
+    def analytic_probability(self, wires=None):
```
antalszava, line 362: This change follows-up the change in the structure of QubitDevice:
XanaduAI/pennylane#573

PR #50, opened 2020-05-13
Incrementing the version number to 0.9.0

antalszava:
- Increments the version number
- Increases the PyQuil requirement to the latest version
- Increases the PennyLane version due to API changes


josh146:
Thanks @antalszava! We'll need to merge this in as soon as PL v0.9 is released and on PyPI

PR #51, opened 2020-05-25
Update from lattice-based to device-based naming

trbromley:
Recent server-side changes on QCS have removed lattice-based devices such as `Aspen-4-5Q-E` and replaced with the full device, simply `Aspen-4`.

This is a breaking change in PL-Forest, which uses the lattice specification to load the number of qubits.

This PR fixes the issue: users can now load a device with
`dev = qml.device("forest.qvm", device="Aspen-4", shots=100)`

### DIFF
--- a/pennylane_forest/device.py
+++ b/pennylane_forest/device.py
@@ -166,17 +166,6 @@ class ForestDevice(QubitDevice):
         shots (int): Number of circuit evaluations/random samples used
             to estimate expectation values of observables.
             For simulator devices, 0 means the exact EV is returned.
-
-    Keyword args:
-        forest_url (str): the Forest URL server. Can also be set by
-            the environment variable ``FOREST_URL``, or in the ``~/.qcs_config``
-            configuration file. Default value is ``"https://forest-server.qcs.rigetti.com"``.
-        qvm_url (str): the QVM server URL. Can also be set by the environment
-            variable ``QVM_URL``, or in the ``~/.forest_config`` configuration file.
-            Default value is ``"http://127.0.0.1:5000"``.
-        quilc_url (str): the compiler server URL. Can also be set by the environment
-            variable ``QUILC_URL``, or in the ``~/.forest_config`` configuration file.
-            Default value is ``"http://127.0.0.1:6000"``.
     """
     pennylane_requires = ">=0.9"
     version = __version__
--- a/pennylane_forest/device.py
+++ b/pennylane_forest/device.py
@@ -188,35 +188,21 @@ class ForestDevice(QubitDevice):
     def __init__(self, wires, shots=1000, analytic=False, **kwargs):
         super().__init__(wires, shots, analytic=analytic)
         self.analytic = analytic
-        self.forest_url = kwargs.get("forest_url", pyquil_config.forest_url)
-        self.qvm_url = kwargs.get("qvm_url", pyquil_config.qvm_url)
-        self.compiler_url = kwargs.get("compiler_url", pyquil_config.quilc_url)
+        self.reset()
 
-        self.connection = ForestConnection(
-            sync_endpoint=self.qvm_url,
-            compiler_endpoint=self.compiler_url,
-            forest_cloud_endpoint=self.forest_url,
+    @staticmethod
+    def _get_connection(**kwargs):
+        forest_url = kwargs.get("forest_url", pyquil_config.forest_url)
+        qvm_url = kwargs.get("qvm_url", pyquil_config.qvm_url)
+        compiler_url = kwargs.get("compiler_url", pyquil_config.quilc_url)
+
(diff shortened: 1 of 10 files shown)


pennylane_forest/device.py
```diff
@@ -188,35 +188,21 @@ class ForestDevice(QubitDevice):
...
         self.analytic = analytic
-        self.forest_url = kwargs.get("forest_url", pyquil_config.forest_url)
-        self.qvm_url = kwargs.get("qvm_url", pyquil_config.qvm_url)
-        self.compiler_url = kwargs.get("compiler_url", pyquil_config.quilc_url)
-
-        self.connection = ForestConnection(
-            sync_endpoint=self.qvm_url,
-            compiler_endpoint=self.compiler_url,
-            forest_cloud_endpoint=self.forest_url,
-        )
```
trbromley, line 199: Creating a connection is now needed _before_ `init` is called, so we make it a static method.

pennylane_forest/qpu.py
```diff
@@ -152,6 +149,10 @@ def __init__(
             if timeout is not None:
                 self.qc.compiler.client.timeout = timeout
 
+        num_wires = len(self.qc.qubits())
+
+        super(QVMDevice, self).__init__(num_wires, shots, **kwargs)
+
```
trbromley, line 155: To explain what's going on: before we could infer the qubit number from the lattice. Now the best way seems to be to load the device and query the number of qubits.

Beforehand, we ran `super(QVMDevice, self).__init__` to create a connection and then loaded the qc. However, we need the number of wires as an argument in `super(QVMDevice, self).__init__`.

Now, we create a connection using the `_get_connection` method, then load the qc, then run `init` with the number of wires evaluated.

pennylane_forest/qvm.py
```diff
@@ -106,28 +106,7 @@ def __init__(self, device, *, shots=1000, noisy=False, **kwargs):
...
-                raise ValueError("QVM device string does not indicate the number of qubits!")
-
-            num_wires = int(wire_match.groups()[0])
-        else:
-            raise ValueError(
-                "Required argument device must be a string corresponding to "
-                "a valid QVM quantum computer, or a NetworkX graph object."
-            )
-
-        super().__init__(num_wires, shots, analytic=analytic, **kwargs)
```
trbromley, line 130: Now that we load the number of wires in a different way, I think all this block does is input validation of `device`. However, this is something that is also covered by `get_qc` so I don't think we need to repeat here.

tests/test_qpu.py
```diff
@@ -17,16 +17,11 @@
...
+pattern = "Aspen-."
 VALID_QPU_LATTICES = [
     qc for qc in pyquil.list_quantum_computers() if "qvm" not in qc and re.match(pattern, qc)
 ]
 
-pattern_four_qubit = "Aspen-.-4Q-."
-VALID_FOUR_QUBIT_LATTICES = [
-    qc for qc in pyquil.list_quantum_computers() if "qvm" not in qc and re.match(pattern_four_qubit, qc)
-]
-
```
trbromley, line 29: I'm having some trouble with the tests:
- getting connection errors when running on `Aspen8`
- things are also a bit slow even on `Aspen4`, maybe because we are loading more qubits than we need to (e.g., maybe we need just 4). One suggestion is to just have:
`VALID_QPU_LATTICES = ["4q-qvm"]`?
This might speed things up and make the tests less dependent on device naming.

pennylane_forest/device.py
josh146: 🤔

pennylane_forest/qpu.py
josh146: That explains it well. Thanks @trbromley!

pennylane_forest/qvm.py
josh146: Agree!

josh146: This actually simplifies the plugin logic significantly

tests/test_qpu.py
josh146: I think this is a good approach, especially as it removes any server-side dependence on the test suite.

tests/test_qvm.py
```diff
@@ -884,7 +864,6 @@ def circuit(params, wires, statement=None):
         for idx, stmt in enumerate(statements):
             qnodes[idx]([], statement=stmt)
             assert dev.circuit_hash in dev._compiled_program_dict
-            print(dev.circuit_hash, dev._compiled_program_dict)
```
josh146, line 887: Good catch!

tests/test_qpu.py
trbromley: Ok will make this change, hopefully then the tests will pass!

PR #53, opened 2020-06-18
Flat tensor wires

antalszava:
Two changes are included:
* A change to how wires are handled for the `Tensor` class, as PennyLane core now stores a flat sequence of wires (as opposed to the previous nested sequence)
* The requirement of PennyLane was changed to the latest master version. This is useful to check that all are fine before the release.

The second change *will have to be reverted* right before the actual release (with the version bump)

### DIFF
--- a/tests/test_qpu.py
+++ b/tests/test_qpu.py
@@ -60,7 +60,7 @@ class TestQPUIntegration(BaseTest):
     @pytest.mark.parametrize(
         "obs", [qml.PauliX(0), qml.PauliZ(0), qml.PauliY(0), qml.Hadamard(0), qml.Identity(0)]
     )
-    def test_tensor_wires_expval_parametric_compilation(self, obs):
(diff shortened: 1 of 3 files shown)


tests/test_qpu.py
```diff
@@ -366,13 +366,14 @@ def circuit():
...
-        assert np.isclose(result, 0.5, atol=2e-2)
+        assert np.isclose(result, 0.5, atol=3e-2)
 
     @flaky(max_runs=5, min_passes=3)
     def test_2q_gate(self):
         """Test that the two qubit gate with the PauliZ observable works correctly.
 
-        As the results coming from the qvm are stochastic, a constraint of 3 out of 5 runs was added.
+        As the results coming from the qvm are stochastic, a constraint of 3
+        out of 5 runs was added.
```
josh146, line 376:
```suggestion
        As the results coming from the qvm are stochastic, a constraint of 3 out of 5 runs was added.
```

PR #55, opened 2020-08-06
[WIRES] Make plugin compatible with wires refactor

mariaschuld:
This PR updates the plugin to support the new wire management in PennyLane master.

Along the way, a few other breaking changes are fixed.

**How wires are now handled**:

* Wires, for example as extracted from ``operator.wires`` are now ``pennylane.wires.Wires`` objects. This class is a subclass of `Sequence`. A tuple of the "labels" (such as `(0, 1, 2)`) can be extracted via `wires.labels`. Lists and arrays can be extracted with the `tolist()` and `toarray()` methods.
* Devices can now be initialised with a Sequence of arbitrary wire labels that the device has to map to its internal labels. This is done with the `map_wires` method.
* A device can overwrite the ``define_wire_map`` function to define a wire map.
* As a guideline, the ``Wires`` objects should only be translated to device wire labels, and unpacked to extract the actual labels as late as possible in the function tree. Exceptions are only where this would lead to unwanted code duplication.


pennylane_forest/device.py
```diff
@@ -33,17 +33,20 @@
...
 
+from collections import OrderedDict
+
 from pyquil import Program
 from pyquil.api._base_connection import ForestConnection
 from pyquil.api._config import PyquilConfig
 
 from pyquil.quil import DefGate
-from pyquil.gates import X, Y, Z, H, PHASE, RX, RY, RZ, CZ, SWAP, CNOT
+from pyquil.gates import X, Y, Z, H, PHASE, RX, RY, RZ, CZ, SWAP, CNOT, S, T, CSWAP
```
mariaschuld, line 43: Just for completeness: these gates are now supported.

pennylane_forest/device.py
```diff
@@ -198,16 +203,15 @@ def program(self):
         """View the last evaluated Quil program"""
         return self.prog
 
-    def remap_wires(self, wires):
-        """Use the wiring specified for the device if applicable.
+    def define_wire_map(self, wires):
```
mariaschuld, line 206: This new function fulfils the same task, but uses different data structures.

pennylane_forest/device.py
```diff
@@ -245,33 +249,34 @@ def apply_rotations(self, rotations):
...
-        self._active_wires = set()
+        self._active_wires = Wires([])
         self._state = None
 
     @property
     def operations(self):
         return set(self._operation_map.keys())
 
-    def mat_vec_product(self, mat, vec, wires):
+    def mat_vec_product(self, mat, vec, device_wire_labels):
```
mariaschuld, line 268: One could also pass the original `wires` object to this method and unpack inside (to follow the guideline of unpacking and translating as late as possible). But here it seemed reasonable to define the function on the device wire labels.

pennylane_forest/numpy_wavefunction.py
```diff
@@ -45,7 +45,7 @@ class NumpyWavefunctionDevice(ForestDevice):
 
     def __init__(self, wires, *, shots=1000, analytic=True, **kwargs):
         super().__init__(wires, shots, analytic, **kwargs)
-        self.qc = PyQVM(n_qubits=wires, quantum_simulator_type=NumpyWavefunctionSimulator)
+        self.qc = PyQVM(n_qubits=len(self.wires), quantum_simulator_type=NumpyWavefunctionSimulator)
```
mariaschuld, line 48: Now `wires` is not necessarily an int any more...

pennylane_forest/wavefunction.py
```diff
@@ -95,17 +96,19 @@ def expand_state(self):
             # all wires in the device have been initialised
             return
 
-        # there are some wires in the device that have not yet been initialised
-        inactive_wires = set(range(self.num_wires)) - self._active_wires
+        num_inactive_wires = len(self.wires) - len(self._active_wires)
```
mariaschuld, line 99: We only need the number of inactive wires here...and the `Wires` objects make sure that labels are unique!

tests/test_converter.py
```diff
@@ -6,7 +6,7 @@
 import pyquil
 import pyquil.gates as g
 import pytest
-from pennylane.utils import OperationRecorder
+from pennylane._queuing import OperationRecorder
```
mariaschuld, line 9: An unrelated breaking change from PL.

mariaschuld: More below :)

tests/test_numpy_wavefunction.py
```diff
@@ -32,7 +32,7 @@ def test_var(self, tol):
         O = qml.var(qml.PauliZ(wires=[0]))
 
         observables = [O]
-        circuit_graph = qml.CircuitGraph(circuit_operations + observables, {})
+        circuit_graph = qml.CircuitGraph(circuit_operations + observables, {}, dev.wires)
```
mariaschuld, line 35: CircuitGraph now takes the wires addressable on a device, to infer ordering...

tests/test_ops.py
```diff
@@ -1,8 +1,9 @@
 import numpy as np
 import pennylane as qml
-import pennylane.plugins.default_qubit as dq
+import pennylane.devices.default_qubit as dq
```
mariaschuld, line 3: Another unrelated breaking change..

tests/test_pyqvm.py
```diff
@@ -2,8 +2,6 @@
 Unit tests for the pyQVM simulator device.
 """
 import logging
-
-import networkx as nx
```
mariaschuld, line 6: Deleted unused imports.

mariaschuld:
I think the checks compare this branch to PL's last stable release, not with master - the errors are all import errors from new modules.

pennylane_forest/device.py
josh146: makes sense!

pennylane_forest/wavefunction.py
josh146: Nice!

tests/test_converter.py
josh146: Thanks for catching this!

PR #58, opened 2020-08-10
Delete superfluous wires conversion

mariaschuld:
This PR fixes a failing test in the shared device test suite by deleting a wrong wire conversion that was actually not required in the first place.

### DIFF
--- a/pennylane_forest/device.py
+++ b/pennylane_forest/device.py
@@ -361,7 +361,5 @@ class ForestDevice(QubitDevice):
         if self._state is None:
             return None
 
-        wires = wires or range(self.num_wires)
-        wires = Wires(wires)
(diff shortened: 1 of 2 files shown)


pennylane_forest/device.py
```diff
@@ -361,7 +361,5 @@ def analytic_probability(self, wires=None):
         if self._state is None:
             return None
 
-        wires = wires or range(self.num_wires)
-        wires = Wires(wires)
```
mariaschuld, line 365: ``marginal_prob`` is written to return ``prob`` when ``wires is None``, so this (wrong) conversion to ``range(self.num_wires)`` is actually not needed at all - we can just feed the wires argument through this function.

mariaschuld:
The failing test might be a stochastic issue, locally the tests pass fine on this branch!

PR #60, opened 2020-08-17
Incrementing the version number to 0.11.0

antalszava:
* [`pennylane-requires` was updated previously](https://github.com/rigetti/pennylane-forest/blob/aed1e59d6f15545b79ff862c85d88260b6c34b7c/pennylane_forest/device.py#L176)


pennylane_forest/_version.py
```diff
@@ -2,4 +2,4 @@
    Version number (major.minor.patch[-label])
```
josh146, line 2:
```suggestion
   Version number (convention major.minor.patch[-label])
```

PR #68, opened 2021-01-28
Tape mode

antalszava:
Prepares the plugin to be compatible with tape mode:

* [X] Updates the gradient tests to use the `requires_grad` argument
* [X] Deprecates the QPU and raises an error
* [X] Disables the converter and raises an error

This PR branches off of #67.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -36,11 +36,6 @@ Currently, PennyLane-Forest provides these Forest devices for PennyLane:
     :description: Forest's QVM and pyQuil pyQVM simulator.
     :link: devices/qvm.html
 
-.. devicegalleryitem::
-    :name: 'forest.qpu'
-    :description: Forest's QPU hardware devices.
-    :link: devices/qpu.html
-
 .. raw:: html
 
         <div style='clear:both'></div>
@@ -79,7 +74,6 @@ Tutorials that originally showcase the forest device are the demos on
    devices/numpy_wavefunction
    devices/wavefunction
    devices/qvm
-   devices/qpu
 
 .. toctree::
    :maxdepth: 1
--- a/pennylane_forest/converter.py
+++ b/pennylane_forest/converter.py
@@ -1,7 +1,7 @@
 import warnings
+from collections.abc import Sequence
 import pkg_resources
 from packaging import version
-from collections.abc import Sequence
 
 import numpy as np
 import pennylane as qml
@@ -329,10 +329,12 @@ class ProgramLoader:
     _matrix_dictionary = pyquil.simulation.matrices.QUANTUM_GATES
 
     def __init__(self, program):
-        pl_version = pkg_resources.get_distribution('pennylane').version
+        pl_version = pkg_resources.get_distribution("pennylane").version
         if version.parse(pl_version) >= version.parse("0.14.0.dev"):
-            raise ValueError("The PyQuil program conversion feature is not \
-                supported with PennyLane version 0.14.0 and higher.")
+            raise ValueError(
+                "The PyQuil program conversion feature is not \
+                supported with PennyLane version 0.14.0 and higher."
+            )
 
         self.program = program  # pylint: disable=unreachable
         self.qubits = program.get_qubits()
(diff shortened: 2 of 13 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
 pyquil>=2.16
 git+https://github.com/PennyLaneAI/pennylane.git
 networkx
-flaky
+flaky
```
antalszava, line 4: newline change

pennylane_forest/converter.py
```diff
@@ -327,7 +327,9 @@ class ProgramLoader:
     _matrix_dictionary = pyquil.simulation.matrices.QUANTUM_GATES
 
     def __init__(self, program):
-        self.program = program
+        raise ValueError("The PyQuil program conversion feature is currently not supported.")
```
josh146, line 330: :+1:

It might be worth removing the device card on the QPU from the docs! and maybe also removing the converter.

Another option could be to simply have a `.. warning::` box that the converter is not support with the latest version of PL

josh146 (approved):
💯 Awesome, looks good from my end Antal!

doc/index.rst
```diff
@@ -36,11 +36,6 @@ Currently, PennyLane-Forest provides these Forest devices for PennyLane:
     :description: Forest's QVM and pyQuil pyQVM simulator.
     :link: devices/qvm.html
 
-.. devicegalleryitem::
```
trbromley, line 39: We also mention QPUs once in the README:
> Contains the PennyLane Forest plugin. This plugin allows different Rigetti devices to work with PennyLane — the wavefunction simulator, the Quantum Virtual Machine (QVM), and Quantum Processing Units (QPUs).

Could just remove the last few words?

pennylane_forest/converter.py
```diff
@@ -585,6 +587,10 @@ def load_program(program: pyquil.Program):
     Every variable that is present in the Program and that is not used as the target
     register of a measurement has to be provided in the ``parameter_map`` of the template.
 
+    .. warning::
+
+        Converting from Forest is not supported with PennyLane version 0.14.0.
```
trbromley, line 592: Given that we raise an error above, is there any version of PL that this would work with?

pennylane_forest/converter.py
trbromley: Could we raise only if version 0.14 or above of PL?

pennylane_forest/qpu.py
```diff
@@ -100,8 +100,9 @@ def __init__(
         calibrate_readout="plus-eig",
         **kwargs,
     ):
+        raise ValueError("Using the QPU via Forest is being deprecated.")
 
-        if readout_error is not None and load_qc:
+        if readout_error is not None and load_qc:  # pylint: disable=unreachable
```
trbromley, line 105: Why is this unreachable? :thinking:

tests/test_qpu.py
```diff
@@ -66,7 +64,6 @@ def test_qpu_args(self):
         with pytest.raises(ValueError, match="Device has a fixed number of"):
             qml.device("forest.qpu", device=device, shots=5, wires=range(100), load_qc=False)
 
-    @flaky(max_runs=10, min_passes=3)
```
trbromley, line 69: Do we need to remove flaky here? Is it to stop 10 repetitions of a test we know will fail?

trbromley (approved):
Thanks @antalszava, looks good to me! Just left some minor comments.

doc/index.rst
antalszava: Good catch, thanks! Updated, also removed the device such that it's not provided by the plugin

pennylane_forest/converter.py
antalszava: Updated the message, it should work with PL version `<0.14.0`

pennylane_forest/qpu.py
antalszava: There is an error that is raised before that

tests/test_qpu.py
antalszava: Yes, that was the original purpose, but rather placed them in a comment now.

pennylane_forest/converter.py
antalszava: Updated to do that 👍

PR #69, opened 2021-02-01
Incrementing the version number to 0.14.0

antalszava:
* The PennyLane version requirement was set back to `pennylane>=0.11` because
    * The converter feature and the `QPUDevice` was disabled based on the PennyLane version (`pennylane>=0.14`)
    * Only the tests of the plugin had to be updated
* Updated the changelog
* Incremented the version number

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,28 @@
-# Release 0.13.0-dev
+# Release 0.14.0
 
 ### New features since last release
 
 ### Improvements
 
+* Updated the CI.
+  [(#63)](https://github.com/PennyLaneAI/pennylane-forest/pull/63)
+
 ### Bug fixes
 
+* Updated the plugin to be compatible with the new core of PennyLane.
+  [(#67)](https://github.com/PennyLaneAI/pennylane-forest/pull/67)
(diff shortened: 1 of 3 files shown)


antalszava:
The tests fail because they were adjusted according to some changes added in `v0.14.0` of PennyLane (e.g., see failures of the `test_ops.py` as per the changes in this PR https://github.com/PennyLaneAI/pennylane-forest/pull/67). Yet there were no changes in the plugin itself that would necessitate increasing the requirement of PennyLane.

@josh146 Would we still want to bump the version of PennyLane?

CHANGELOG.md
```diff
@@ -1,17 +1,29 @@
...
 ### Documentation
 
+* Adapted the documentation to the PennyLane theme.
+  [(#64)](https://github.com/PennyLaneAI/pennylane-forest/pull/64)
+
 ### Contributors
 
 This release contains contributions from (in alphabetical order):
 
+Thomas Bromley, Theodor Isacsson, Josh Izaac, Maria Schuld, Antal Száva.
```
josh146, line 25: Wow, most contributors ever!

josh146 (approved):
@antalszava if only the tests were updated, no need to bump the PL requirements!

PR #71, opened 2021-04-08
Pre v0 15 fixes

antalszava:
- [X] Removes the use of the `analytic` keyword to `shots=None`,
- [X] Changes from checking for a `Variable` to a condition on `requires_grad`,
- [X] Adds the circuit hash computation feature required by parametric compilation (as per https://github.com/PennyLaneAI/pennylane/pull/1182).
- [x] Refactor device tests (failing on the `CircuitGraph` object which is passed to apply)

Note: disallowed `BasisState` to be used for parametric compilation, as internally we don't use any parametrized gates to achieve it's effect (but rather apply PauliX and Identity operations, see `basis_state` function in `device.py`).

### DIFF
--- a/pennylane_forest/qvm.py
+++ b/pennylane_forest/qvm.py
@@ -50,8 +50,10 @@ class QVMDevice(ForestDevice):
             * Any other supported Rigetti device architecture.
             * Graph topology representing the device architecture.
 
-        shots (int): number of circuit evaluations/random samples used
-            to estimate expectation values of observables.
+        shots (None, int, list[int]): Number of circuit evaluations/random samples used to estimate
+            expectation values of observables. If ``None``, the device calculates probability, expectation values,
+            and variances analytically. If an integer, it specifies the number of samples to estimate these quantities.
+            If a list of integers is passed, the circuit evaluations are batched over the list of shots.
         wires (Iterable[Number, str]): Iterable that contains unique labels for the
             qubits as numbers or strings (i.e., ``['q1', ..., 'qN']``).
             The number of labels must match the number of qubits accessible on the backend.
@@ -79,10 +81,9 @@ class QVMDevice(ForestDevice):
 
     def __init__(self, device, *, wires=None, shots=1000, noisy=False, **kwargs):
 
-        if shots <= 0:
-            raise ValueError("Number of shots must be a positive integer.")
+        if shots is not None and shots <= 0:
+            raise ValueError("Number of shots must be a positive integer or None.")
 
-        analytic = kwargs.get("analytic", False)
         timeout = kwargs.pop("timeout", None)
 
         self._compiled_program = None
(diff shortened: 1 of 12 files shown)


pennylane_forest/qvm.py
```diff
@@ -192,10 +203,10 @@ def apply_parametric_program(self, operations, **kwargs):
             # Prepare for parametric compilation
             par = []
             for param in operation.data:
-                if isinstance(param, Variable):
-                    # Using the idx for each Variable instance to specify the
+                if getattr(param, "requires_grad", False) and operation.name is not "BasisState":
```
antalszava, line 206:
```suggestion
                if getattr(param, "requires_grad", False) and operation.name != "BasisState":
```

pennylane_forest/qvm.py
```diff
@@ -192,10 +203,10 @@ def apply_parametric_program(self, operations, **kwargs):
             # Prepare for parametric compilation
             par = []
             for param in operation.data:
-                if isinstance(param, Variable):
-                    # Using the idx for each Variable instance to specify the
+                if getattr(param, "requires_grad", False) and operation.name != "BasisState":
```
antalszava, line 206: The reason `BasisState` is disallowed, is we have a custom `basis_state` function in `device.py` that applies PauliX and Identity operations based on the state that is provided to `BasisState`.

It is unnecessary (and yields incorrect results) to then define a PyQuil memory reference for the array of states provided to `BasisState`, since we'd not have parametrized operations (only PauliX and Identity).

antalszava:
The codefactor errors are out of scope, shall we ignore them? (either inherited docstring from PL core or unused `**kwargs`)

requirements.txt
```diff
@@ -1,4 +1,4 @@
 pyquil>=2.16
-pennylane>=0.11
+git+https://github.com/PennyLaneAI/pennylane.git
```
antalszava, line 2:
```suggestion
git+https://github.com/PennyLaneAI/pennylane.git@master
```

setup.py
```diff
@@ -7,7 +7,7 @@
     version = f.readlines()[-1].split()[-1].strip("\"'")
 
 
-requirements = ["pyquil>=2.16", "pennylane>=0.11"]
+requirements = ["pyquil>=2.16", "pennylane @ git+https://github.com/PennyLaneAI/pennylane.git"]
```
antalszava, line 10:
```suggestion
requirements = ["pyquil>=2.16", "pennylane @ git+https://github.com/PennyLaneAI/pennylane.git@master"]
```

pennylane_forest/qvm.py
josh146: 👍

josh146 (approved):
Thanks so much for this big effort @antalszava, and fantastic to see parametric compilation still working!

PR #77, opened 2021-08-04
Temporarily pin `pyquil>=2.16,<2.28.3`

antalszava:


rmoyard (commented):
It is a good temporary fix 💯  Maybe you can add it in the changelog?

josh146:
Thanks @antalszava and @rmoyard! It might also be worth updating the readme to note this, to make it more visible. Something that comes to mind -- perhaps there are users that need the PyQuil v3.0 release to target the new APIs, so it would be good to make it explicit we don't support this (for now).

antalszava:
[ch7572]

PR #80, opened 2021-08-09
Increment version number to v0.17.0 and update QVM tests

antalszava:
* Bumps the version number to v0.17.0
* Changes QVM tests:

Sampling has been changed in
https://github.com/PennyLaneAI/pennylane/commit/2344a4acfb9698e26af79cbe67195c7bbe859922.

This affects how current QVM tests are handled. The tests create a `MeasurementProcess` and this affects the sampling mechanism.

### DIFF
--- a/tests/test_qvm.py
+++ b/tests/test_qvm.py
@@ -58,8 +58,8 @@ class TestQVMBasic(BaseTest):
             qml.RX(theta, wires=[0])
             qml.RX(phi, wires=[1])
             qml.CNOT(wires=[0, 1])
-            O1 = qml.Identity(wires=[0])
-            O2 = qml.Identity(wires=[1])
+            O1 = qml.expval(qml.Identity(wires=[0]))
+            O2 = qml.expval(qml.Identity(wires=[1]))
 
         dev.apply(tape.operations, rotations=tape.diagonalizing_gates)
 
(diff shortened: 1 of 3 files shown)


tests/test_qvm.py
```diff
@@ -65,7 +65,7 @@ def test_identity_expectation(self, shots, qvm, compiler):
 
         dev._samples = dev.generate_samples()
 
-        res = np.array([dev.expval(O1), dev.expval(O2)])
+        res = np.array([dev.expval(O1.obs), dev.expval(O2.obs)])
```
thisac, line 68: Was this somehow changed to not support having expvals called twice (thus the need to extract the identity here before calling `dev.expval`?

thisac (approved):
💯

tests/test_qvm.py
antalszava: Hmm, not sure I understand :thinking: The number of calls to `dev.expval` should not matter, we just need to extract the observable so that we're passing a `qml.operation.Observable` rather than a `MeasurementProcess`.

thisac: Oh, I guess I'm simply wondering why it seemingly was working before (or maybe it wasn't) and why this change is needed. It seems like an obvious change.

antalszava:
[ch7894]

PR #83, opened 2021-08-13
Update setup.py

antalszava:
Same as #82 just for `setup.py`.


rmoyard (approved):
Thanks, it looks good to me 💯

PR #84, opened 2021-08-13
Use `@master` when pinning the PL requirement

antalszava:
As the plugin test matrix explicitly uses the `@master` pinning, we'd like to pin accordingly here too. Other `pip` has issues with the resolution.


rmoyard (approved):
Perfect, it looks good to me! 💯

PR #88, opened 2021-11-30
Fix test because of `par_domain` removal

rmoyard:
Change one test due to the removal of `par_domain` and `num_param` attributes in the PennyLane operator class.

### DIFF
--- a/tests/test_numpy_wavefunction.py
+++ b/tests/test_numpy_wavefunction.py
@@ -69,69 +69,60 @@ class TestWavefunctionBasic(BaseTest):
 
         self.assertAlmostEqual(var, expected, delta=tol)
 
-    @pytest.mark.parametrize(
-        "gate", plf.NumpyWavefunctionDevice._operation_map
-    )  # pylint: disable=protected-access
-    def test_apply(self, gate, apply_unitary, tol):
+    def test_apply(self, apply_unitary, tol):
         """Test the application of gates to a state"""
         dev = plf.NumpyWavefunctionDevice(wires=3)
 
-        try:
-            # get the equivalent pennylane operation class
-            op = getattr(qml.ops, gate)
-        except AttributeError:
-            # get the equivalent pennylane-forest operation class
-            op = getattr(plf, gate)
-
-        # the list of wires to apply the operation to
-        w = list(range(op.num_wires))
-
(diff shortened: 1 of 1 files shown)


tests/test_numpy_wavefunction.py
```diff
@@ -402,5 +393,4 @@ def circuit(x, y, z):
             runs.append(circuit(a, b, c))
 
         expected_var = np.sqrt(1 / shots)
-        print(np.mean(runs), np.cos(a) * np.sin(b))
```
antalszava, line 405: Nice catch!

tests/test_numpy_wavefunction.py
```diff
@@ -69,69 +69,60 @@ def test_var_hermitian(self, tol):
...
-
-        dev.apply(tape.operations, rotations=tape.diagonalizing_gates)
 
-        res = dev.expval(obs)
+            dev.apply(tape.operations, rotations=tape.diagonalizing_gates)
 
-        # verify the device is now in the expected state
-        self.assertAllAlmostEqual(dev._state, state, delta=tol)
+            # verify the device is now in the expected state
+            self.assertAllAlmostEqual(dev._state, state, delta=tol)
```
antalszava, line 125: Could these statements be executed outside of the context?

tests/test_numpy_wavefunction.py
```diff
@@ -69,69 +69,60 @@ def test_var_hermitian(self, tol):
...
-        w = list(range(op.num_wires))
-
         obs = qml.expval(qml.PauliZ(0))
-        if op.par_domain == "A":
-            # the parameter is an array
-            if gate == "QubitUnitary":
-                p = np.array(U)
-                w = [0]
+
+        ops = [
```
antalszava, line 78: The previous version seems to have been using parametrization. It would be good to keep that style because that way we can be sure which gate application fails if the test fails.

antalszava (changes requested):
Hi @rmoyard, thank you so much for fixing this! :tada: One change I think would be good to have: the test case seems to have been parametrizing over gates previously. It seems like it could be good to keep that structure.

antalszava (approved):
Looks good! :tada: Thank you :slightly_smiling_face:

PR #90, opened 2021-12-07
Fix integration tests and adapt test for `par_domain`removal

rmoyard:
Due to recent PennyLane changes, some parameters are given as array and it is not supported by Pyquil. We apply a small fix to transform the parameters to floats.

Multiple tests were also changed due to the removal of `par_domain` in PennyLane.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ with open("pennylane_forest/_version.py") as f:
     version = f.readlines()[-1].split()[-1].strip("\"'")
 
 
-requirements = ["pyquil>=2.16,<2.28.3", "pennylane>=0.17"]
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -7,7 +7,7 @@
     version = f.readlines()[-1].split()[-1].strip("\"'")
 
 
-requirements = ["pyquil>=2.16,<2.28.3", "pennylane>=0.17"]
+requirements = ["pyquil>=2.16,<2.28.3", "pennylane @ git+https://github.com/PennyLaneAI/pennylane.git@master"]
```
antalszava, line 10: Not pinning the branch should potentially help with getting the demos to run.
```suggestion
requirements = ["pyquil>=2.16,<2.28.3", "pennylane @ git+https://github.com/PennyLaneAI/pennylane.git"]
```

antalszava (approved):
Looks good to me! :tada: As suggested, one step will be the changelog.

PR #99, opened 2022-03-22
Fix docs links to point to `PennyLaneAI` organization

antalszava:
Some links of PennyLane and the PennyLane plugin were still pointing to the `XanaduAI` organization instead of `PennyLaneAI`. These are now updated and the link for the blog points to the PennyLane blog.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -271,11 +271,11 @@ html_theme_options = {
     "table_header_border": "#19b37b",
     "download_button": "#19b37b",
     # gallery options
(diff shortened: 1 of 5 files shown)


albi3ro:
@antalszava While we're checking links, the link to the "Ensemble classification with Forest and Qiskit devices" is broken.

albi3ro:
Also at the bottom of the page under "About", The link for "Software" is broken.

antalszava:
Thanks for the catches @albi3ro! :tada:

albi3ro (approved):
Thanks for the update.

PR #101, opened 2022-06-08
Define `state` method and `returns_state=True` in `capabilities`

antalszava:
The Wavefunction and NumPyWavefunction simulator devices are capable of returning the state, yet they don't define a `state` method or have the `returns_state=True` entry in the `capabilities` dictionary of the device (see [the explanation of `capabilities` here](https://pennylane.readthedocs.io/en/stable/development/plugins.html#device-capabilities)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Release 0.21.0-dev
+# Release 0.24.0-dev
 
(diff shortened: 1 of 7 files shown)


rmoyard (approved):
Thank you again @antalszava, it looks good!

Issue #45, opened 2020-02-11
Remove sample collection redundancy in `QPUDevice.expval`


msohaibalam:
Currently, when using the `QPUDevice.expval` method, both pyQuil's `measure_observables` as well as PennyLane's `generate_samples` methods are called. This leads to redundancy, as both methods generate samples, while only those from `measure_observables` are actually (currently) used in the estimation of the expectation value. See the discussion at https://github.com/rigetti/pennylane-forest/pull/42#discussion_r376502068.

We could either over-ride `generate_samples` when using `measure_observables`, or try to use the samples from `generate_samples` and the error mitigation routines from `measure_observables`.

mariaschuld:
Closing for now, feel free to re-open.

antalszava:
This issue is still relevant.

antalszava:
Note that this is a **major** shortcoming of using the QPU device with operator estimation. Users will see an increased number of device execution in such cases that could be significantly decreased.

PR #107, opened 2022-10-04
Migrate to pyQuil v3

MarquessV:
Here are the changes required to get this plugin on to pyQuil v3! I'm including some of my own targeted questions as comments below.


README.rst
```diff
@@ -28,11 +28,11 @@ PennyLane Forest Plugin
 .. header-start-inclusion-marker-do-not-remove
 
 The PennyLane Forest plugin allows different Rigetti devices to work with
-PennyLane --- the wavefunction simulator, and the Quantum Virtual Machine (QVM).
+PennyLane --- the wavefunction simulator, and the Quantum Virtual Machine (QVM), and Quantum Processing Units (QPUs).
```
AlbertMitjans, line 31:
```suggestion
PennyLane --- the wavefunction simulator, the Quantum Virtual Machine (QVM), and Quantum Processing Units (QPUs).
```

README.rst
```diff
@@ -47,8 +47,8 @@ Features
 ========
 
 * Provides three devices to be used with PennyLane: ``forest.numpy_wavefunction``,
```
AlbertMitjans, line 49:
```suggestion
* Provides four devices to be used with PennyLane: ``forest.numpy_wavefunction``,
```

README.rst
```diff
@@ -47,8 +47,8 @@ Features
...
-  ``forest.wavefunction``, and ``forest.qvm``. These provide access to the pyQVM
-  Numpy wavefunction simulator, Forest wavefunction simulator, and quantum
+  ``forest.wavefunction``, ``forest.qvm``, and ``forest.qpu``. These provide access to the pyQVM
+  Numpy wavefunction simulator, pyQuil wavefunction simulator, and quantum
   virtual machine (QVM) respectively.
```
AlbertMitjans, line 52:
```suggestion
  Numpy wavefunction simulator, pyQuil wavefunction simulator, quantum
  virtual machine (QVM), and quantum processing units (QPUs) respectively.
```

README.rst
```diff
@@ -59,7 +59,7 @@ Features
   ``ISWAP``, ``PSWAP``, and ``CPHASE``. Every custom operation supports analytic
   differentiation.
```
AlbertMitjans, line 60: These gates are now defined in PennyLane:
- `ISWAP`: 
https://github.com/PennyLaneAI/pennylane/blob/e79f264adf743dbd35f2811de367edb9789718c3/pennylane/ops/qubit/non_parametric_ops.py#L1426

- `PSWAP`: https://github.com/PennyLaneAI/pennylane/blob/e79f264adf743dbd35f2811de367edb9789718c3/pennylane/ops/qubit/parametric_ops.py#L3189

- `CPHASE`: 
https://github.com/PennyLaneAI/pennylane/blob/e79f264adf743dbd35f2811de367edb9789718c3/pennylane/ops/qubit/parametric_ops.py#L546

README.rst
```diff
@@ -86,14 +86,12 @@ Dependencies
 
 PennyLane-Forest requires the following libraries be installed:
 
-* `Python <http://python.org/>`__ >=3.6
+* `Python <http://python.org/>`__ >=3.7
 
 as well as the following Python packages:
 
 * `PennyLane <http://pennylane.readthedocs.io/>`__
```
AlbertMitjans, line 93: We may need to add a lower bound on the PennyLane version. Will look more into it.

AlbertMitjans (commented):
Left some comments on the `README.rst` file

doc/devices/qpu.rst
```diff
@@ -1,5 +1,3 @@
-:orphan:
-
 The QPU device
```
AlbertMitjans, line 1: Delete duplicated `very similar` text in the first paragraph of this section

doc/devices/qpu.rst
```diff
@@ -64,33 +62,18 @@ These operations include:
     pennylane_forest.ops.ISWAP
     pennylane_forest.ops.PSWAP
```
AlbertMitjans, line 63: The three gates mentioned here are now supported in PennyLane. Maybe remove the last part of this section?

doc/devices/qpu.rst
```diff
@@ -64,33 +62,18 @@ These operations include:
...
-
-``qvm_url`` (*str*)
-    the QVM server URL. Can also be set by the environment
-    variable ``QVM_URL``, or in the ``~/.forest_config`` configuration file.
-    Default value is ``"http://127.0.0.1:5000"``.
-
-``compiler_url`` (*str*)
-    the compiler server URL. Can also be set by the environment
-    variable ``COMPILER_URL``, or in the ``~/.forest_config`` configuration file.
-    Default value is ``"http://127.0.0.1:6000"``.
```
AlbertMitjans, line 86: Why did you delete these keyword arguments? I believe you talk about them in the `quilc server configuration` right below.

doc/devices/qvm.rst
```diff
@@ -190,36 +190,4 @@ After measuring the qubit state, we can determine the probability :math:`P_0` of
```
AlbertMitjans, line 190: Consider removing the `CPHASE`, `ISWAP` and `PSWAP` gates from the `Supported operations` section of this file, given that they are already supported in PennyLane.

doc/devices/qvm.rst
```diff
@@ -190,36 +190,4 @@ After measuring the qubit state, we can determine the probability :math:`P_0` of
...
-
-``qvm_url`` (*str*)
-    the QVM server URL. Can also be set by the environment
-    variable ``QVM_URL``, or in the ``~/.forest_config`` configuration file.
-    Default value is ``"http://127.0.0.1:5000"``.
-
-``compiler_url`` (*str*)
-    the compiler server URL. Can also be set by the environment
-    variable ``COMPILER_URL``, or in the ``~/.forest_config`` configuration file.
-    Default value is ``"http://127.0.0.1:6000"``.
```
AlbertMitjans, line 215: Same as before, you delete these keyword arguments but then mention them in the next section.

doc/devices/wavefunction.rst
```diff
@@ -50,37 +50,6 @@ You can then execute the circuit like any other function to get the quantum mech
...
-
-``qvm_url`` (*str*)
-    the QVM server URL. Can also be set by the environment
-    variable ``QVM_URL``, or in the ``~/.forest_config`` configuration file.
-    Default value is ``"http://127.0.0.1:5000"``.
-
-``compiler_url`` (*str*)
-    the compiler server URL. Can also be set by the environment
-    variable ``COMPILER_URL``, or in the ``~/.forest_config`` configuration file.
-    Default value is ``"http://127.0.0.1:6000"``.
```
AlbertMitjans, line 72: Same as before.

doc/devices/wavefunction.rst
```diff
@@ -95,4 +64,6 @@ These operations include:
 .. autosummary::
     pennylane_forest.ops.CPHASE
     pennylane_forest.ops.ISWAP
-    pennylane_forest.ops.PSWAP
\ No newline at end of file
+    pennylane_forest.ops.PSWAP
```
AlbertMitjans: Same: these operations are now supported in PennyLane

AlbertMitjans: Consider deleting these ops also in the `numpy-wavefunction` section

doc/index.rst
```diff
@@ -45,7 +50,7 @@ Check out these demos to see the PennyLane-Forest plugin in action:
     <div class="row">
 
 .. title-card::
```
AlbertMitjans, line 52: I think this `title-card` and the next one are not rendered properly (see [link](https://xanaduai-pennylane--107.com.readthedocs.build/projects/forest/en/107/index.html)).

AlbertMitjans (commented):
Added more comments on the `doc` files

pennylane_forest/qpu.py
```diff
@@ -66,28 +61,19 @@ class QPUDevice(QVMDevice):
...
-            the environment variable ``FOREST_SERVER_URL``, or in the ``~/.qcs_config``
-            configuration file. Default value is ``"https://forest-server.qcs.rigetti.com"``.
-        qvm_url (str): the QVM server URL. Can also be set by the environment
-            variable ``QVM_URL``, or in the ``~/.forest_config`` configuration file.
-            Default value is ``"http://127.0.0.1:5000"``.
-        compiler_url (str): the compiler server URL. Can also be set by the environment
-            variable ``COMPILER_URL``, or in the ``~/.forest_config`` configuration file.
-            Default value is ``"http://127.0.0.1:6000"``.
-        timeout (int): number of seconds to wait for a response from the client.
         parametric_compilation (bool): a boolean value of whether or not to use parametric
```
AlbertMitjans, line 70: Should we add `compiler_timeout` and `execution_timeout`?

pennylane_forest/qvm.py
```diff
@@ -228,14 +236,19 @@ def apply_parametric_program(self, operations, **kwargs):
...
-            return self.qc.run(self.prog, memory_map=self._parameter_map)
+            return self.extract_samples(self.qc.run(self.prog))
 
         if self.circuit_hash is None:
             # Parametric compilation was set to False
             # Compile the program
             self._compiled_program = self.qc.compile(self.prog)
-            return self.qc.run(executable=self._compiled_program)
+            results = self.qc.run(executable=self._compiled_program)
+            return self.extract_samples(results)
```
AlbertMitjans, line 251: Careful because in this case the compiled program is not saved in the `self._compiled_program_dict` dictionary.

pennylane_forest/qvm.py
```diff
@@ -244,8 +257,11 @@ def generate_samples(self):
 
         # The program has been compiled, store as the latest compiled program
         self._compiled_program = self._compiled_program_dict[self.circuit_hash]
-        samples = self.qc.run(executable=self._compiled_program, memory_map=self._parameter_map)
-        return samples
+        results = self.qc.run(executable=self._compiled_program)
+        return self.extract_samples(results)
+
+    def extract_samples(self, execution_results):
```
AlbertMitjans, line 263: Add docstring

AlbertMitjans (commented):
Left some comments in the code. Still need to look at the tests.

pennylane_forest/qpu.py
```diff
@@ -19,29 +19,24 @@
...
+from pyquil.paulis import sI, sZ
 from pyquil.quil import Program
 from pyquil.quilbase import Gate
 
 from pennylane.operation import Tensor
 
-from typing import List
-
-from ._version import __version__
 from .qvm import QVMDevice
```
AlbertMitjans, line 40: It is weird that the `QPUDevice` inherits from the `QVMDevice`... In the future we could either move the common methods to the `ForestDevice` class or create a new class

AlbertMitjans:
The `pennylane_forest/device.py` file needs reformatting.

Please run `black -l 100 pennylane_forest/`.

README.rst
MarquessV: Should I just go ahead and remove this whole bullet point then?

doc/devices/qpu.rst
MarquessV: Good catch. These kwargs were removed but I didn't update the note to properly reflect that. I've corrected the note in both places.

doc/index.rst
MarquessV: Looks like the `index-card` directive no longer supports figures. None of the [available directives in the xanadu-sphinx-theme](https://github.com/XanaduAI/xanadu-sphinx-theme#directives) are a perfect match for what I would imagine was being rendered before. There is a reference to a "Related Demo" directive, but it doesn't look like that's been finished yet. What would you recommend?

pennylane_forest/qpu.py
MarquessV: Good catch.

pennylane_forest/qvm.py
MarquessV: That is how it was before too. It looks like the original intent was to not cache the compiled programs if parametric compilation was disabled.

The [`circuit_hash` property returns `None` if parametric compilation is disabled](https://github.com/PennyLaneAI/pennylane-forest/pull/107/files/788b3cc2f73e0bf17700882328200802d89e644b#diff-e54edc4229030c98c5b70918872608e43667b52ebfafa4f6c3d93996eea94466R267).

and the [`_compiled_program_dict` is only initialized if parametric compilation is enabled](https://github.com/PennyLaneAI/pennylane-forest/pull/107/files/788b3cc2f73e0bf17700882328200802d89e644b#diff-e54edc4229030c98c5b70918872608e43667b52ebfafa4f6c3d93996eea94466R89).

Happy to update it if that's not what we want anymore, but want to make sure we're on the same page.

pennylane_forest/device.py
```diff
@@ -176,18 +172,8 @@ def __init__(self, wires, shots=1000, **kwargs):
         self.reset()
```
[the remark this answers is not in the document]
MarquessV: done

pennylane_forest/wavefunction.py
```diff
@@ -72,8 +61,7 @@ class WavefunctionDevice(ForestDevice):
 
     def __init__(self, wires, *, shots=None, **kwargs):
```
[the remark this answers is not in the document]
MarquessV: done - removed it from the other child classes of `ForestDevice` as well.

pennylane_forest/qpu.py
MarquessV: Agreed, that caught me off guard as well. I can create a followup issue if you'd like.

pennylane_forest/qpu.py
```diff
@@ -246,17 +221,19 @@ def expval(self, observable):
                         label, p00=self.readout_error[0], p11=self.readout_error[1]
                     )
 
+            prep_prog.wrap_in_numshots_loop(self.shots)
```
MarquessV, line 224: @AlbertMitjans RE: https://github.com/PennyLaneAI/pennylane-forest/pull/107#discussion_r994449879

Where my confusion came from was that I wasn't sure how this code was working before. The `Program` wasn't being wrapped in a numshots loop and the number of shots wasn't being set on the Experiment either. Upon further investigation it turns out in `pyQuil` v2 an `Experiment` would default to 10k shots if it wasn't set on the `Program` or the `Experiment`.

Before, the code wasn't respecting the number of shots set by the user but it still passed all the tests because it defaulted to 10k shots. Since `Experiment`s no longer set shots in `pyQuil` v3 the program was only running a single time, causing the tests to fail due to inaccuracy and revealing the bug 🎊.

Given this, I removed the comment.

pennylane_forest/qvm.py
AlbertMitjans: Oops my bad, you're right. Given that we are using the `circuit_hash` as dictionary keys, we can't cache it when it is None.

pennylane_forest/device.py
```diff
@@ -36,22 +36,37 @@
...
 from pyquil import Program
-from pyquil.api._base_connection import ForestConnection
-from pyquil.api._config import PyquilConfig
-
 from pyquil.quil import DefGate
-from pyquil.gates import X, Y, Z, H, PHASE, RX, RY, RZ, CZ, SWAP, CNOT, S, T, CSWAP, I
-
-# following gates are not supported by PennyLane
-from pyquil.gates import CPHASE00, CPHASE01, CPHASE10, CPHASE, CCNOT, ISWAP, PSWAP
+from pyquil.gates import (
```
AlbertMitjans, line 40: Missing `CPHASE` import

doc/index.rst
AlbertMitjans: I would just recommend removing the figure. But maybe @josh146 has a better solution.

README.rst
AlbertMitjans: Sure!

pennylane_forest/qpu.py
```diff
@@ -66,28 +61,21 @@ class QPUDevice(QVMDevice):
...
+        execution_timeout (int): number of seconds to wait for a response from the QVM (default 10).
         parametric_compilation (bool): a boolean value of whether or not to use parametric
             compilation.
     """
     name = "Forest QPU Device"
     short_name = "forest.qpu"
     observables = {"PauliX", "PauliY", "PauliZ", "Identity", "Hadamard", "Hermitian"}
+    pennylane_requires = ">=0.15.0"
 
     def __init__(
```
AlbertMitjans, line 80: The `QPUDevice.__init__` contains duplicated code from the `QVMDevice.__init__`

AlbertMitjans: When calling super, use:
```python
super().__init__(wires, shots, **kwargs)
```

Instead of:
```python
super(QVMDevice, self).__init__(wires, shots, **kwargs)
```

AlbertMitjans:
Remove `pytest.mark.xfail` from the `converter.py` tests.

AlbertMitjans:
Do you know if I can run the tests locally on a M1 Mac? The docker image's platform is for linux and is not compatible with mac.

doc/index.rst
MarquessV: I went ahead with that for now (and added a description). Happy to change it later if Josh has a suggestion!

pennylane_forest/device.py
MarquessV: Added it in

pennylane_forest/qpu.py
MarquessV: The way `super` is being used in `QPUDevice` is related to why the code is duplicated between `QPUDevice.__init__` and `QVMDevice.__init__`. 

For example, `QPUDevice` and `QVMDevice` initialize `self.qc` differently. If `QPUDevice` were to call `QVMDevice.__init__`, then it would effectively clobber the `qc` it had setup for itself. Because of this, it bypasses the `QVMDevice` by calling it's superclasses initializer (`ForestDevice.__init__`):

```python
super(QVMDevice, self).__init__(wires, shots, **kwargs)
```

Since they can't share an initializer, `QPUDevice` also has to duplicate all the stuff about initialization it shares in common with `QVMDevice`

As discussed in another thread, this is why `QPUDevice` shouldn't inherit from `QVMDevice`. The right move here is to refactor the shared logic into `ForestDevice` that both devices inherit from. I've gone ahead and started down that path since it's likely to cause more confusion until it's cleaned up.

tests/test_converter.py
```diff
@@ -10,7 +10,6 @@
 from pennylane_forest.converter import *
 
 
-@pytest.mark.xfail
```
MarquessV, line 13: The majority of the converter tests pass, with the exception of 6, each of which fail with:

```
AttributeError: 'QNode' object has no attribute 'jacobian'
```

One example of this is on line 1233:

```
assert np.array_equal(qml.jacobian([angle]), circuit_reg.jacobian([angle]))
```

This seems like a change on the PennyLane side, any idea on what we need to do to fix?

pennylane_forest/qpu.py
AlbertMitjans: Thanks for the explanation! I didn't realise we were calling `super(QVMDevice)` instead of `super(QPUDevice)`.

AlbertMitjans: Then we can also remove the `kwargs` from here 😄

tests/test_converter.py
AlbertMitjans: Yes!

You need to use:
```python
assert np.array_equal(qml.jacobian(circuit)(angle), qml.jacobian(circuit_reg)(angle))
```

Instead of:
```python
assert np.array_equal(qml.jacobian([angle]), circuit_reg.jacobian([angle]))
```

MarquessV: That did the trick, thanks!

pennylane_forest/qc.py
```diff
@@ -0,0 +1,407 @@
...
+        if self.parametric_compilation:
+            self._circuit_hash = circuit.graph.hash
+        return super().execute(circuit, **kwargs)
+
+    def generate_samples(self):
+        if self.parametric_compilation:
+            for region, value in self._parameter_map.items():
+                self.prog.write_memory(region_name=region, value=value)
+
+        if "pyqvm" in self.qc.name:
```
MarquessV, line 255: This might not be necessary anymore, but I want to be able to run the pyqvm test suite before deleting it (along with moving the special case of `pyqvm`s in `apply` to the `QVMDevice`)

pennylane_forest/qpu.py
```diff
@@ -190,7 +163,7 @@ def __init__(
         self.calibrate_readout = calibrate_readout
         self.wiring = {i: q for i, q in enumerate(self.qc.qubits())}
 
-    def expval(self, observable):
+    def expval(self, observable, shot_range=None, bin_size=None):
```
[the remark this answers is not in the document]
MarquessV: I took a brief look at doing this while refactoring the devices, but quickly ran into a wall when I realized that the logic in `expval` is dependent on the `observable` parameter to generate the right `PauliTerm` for the `Experiment` passed to `measure_observables`.

AlbertMitjans:
I just realised that the github workflows run the tests using the `quilc` version 1.22.0. 

Probably that is why some of the tests are passing locally but failing here.

I think we should change the following line in `.github/workflows/tests.yml` from:
```
      - name: Run Forest Quilc
        run: docker run --rm -d -p 5555:5555 rigetti/quilc:1.22.0 -R
```
to
```
      - name: Run Forest Quilc
        run: docker run --rm -d -p 5555:5555 rigetti/quilc -S
```

AlbertMitjans:
Update minimum pennylane version to `0.18` (ISWAP gate was added in the 0.18 release)

setup.py
```diff
@@ -7,7 +7,7 @@
     version = f.readlines()[-1].split()[-1].strip("\"'")
 
 
-requirements = ["pyquil>=2.16,<2.28.3", "pennylane>=0.17"]
+requirements = ["pyquil>=3.3.1,<4.0.0", "qcs-api-client>=0.20.13<0.22.0", "pennylane>=0.17"]
```
AlbertMitjans, line 10:
```suggestion
requirements = ["pyquil>=3.3.1,<4.0.0", "qcs-api-client>=0.20.13<0.22.0", "pennylane>=0.18"]

```

requirements.txt
```diff
@@ -1,4 +1,5 @@
-pyquil>=2.16,<2.28.3
+pyquil>=3.0.0,<4.0.0
 pennylane>=0.17
```
AlbertMitjans, line 2:
```suggestion
pennylane>=0.18
```

AlbertMitjans (commented):
Update minimum pennylane version to `0.18` (ISWAP gate was added in the 0.18 release)

pennylane_forest/converter.py
```diff
@@ -1,7 +1,5 @@
 import warnings
 from collections.abc import Sequence
-import pkg_resources
-from packaging import version
 
 import numpy as np
 import pennylane as qml
```
AlbertMitjans, line 5: Update `pyquil_inv_operation_map` dictionary with:
```python
{
  "S": qml.S,
  "T": qml.T,
  "CCNOT": qml.Toffoli,
  "CSWAP": qml.CSWAP,
  "ISWAP": qml.ISWAP,
  "PSWAP": qml.PSWAP
}
```

And remove `S, T, CCNOT, CSWAP, ISWAP, PSWAP` gates from `pennylane_forest/ops.py`. Leave the `CPHASE` gate because I realised it has a bit more functionality than the `qml.ControlledPhase` gate in PennyLane.

pennylane_forest/converter.py
```diff
@@ -1,7 +1,5 @@
 import warnings
```
AlbertMitjans, line 1: In this file there is a line which uses the `pyquil.simulation` submodule:
```python
_matrix_dictionary = pyquil.simulation.matrices.QUANTUM_GATES
```

For some reason this now crashes the tests. I realised that in the current pyquil version, `simulation` is not imported top level, that is why Python can't find it.

This can be solved by importing simulation directly:
```python
import pyquil.simulation as simulation
```

pennylane_forest/converter.py
AlbertMitjans: The `plf.ops.CCNOT`, `plf.ops.CSWAP`... gates are used more than once in this file.

They all need to be replaced by `qml.Toffoli`, `qml.CSWAP`...

pennylane_forest/qc.py
```diff
@@ -0,0 +1,296 @@
...
+            Setting this to ``True`` results in a significantly faster expectation value
+            evaluation when the number of shots is larger than ~1000.
+
+    Keyword args:
+        compiler_timeout (int): number of seconds to wait for a response from quilc (default 10).
+        execution_timeout (int): number of seconds to wait for a response from the QVM (default 10).
+        parametric_compilation (bool): a boolean value of whether or not to use parametric
+            compilation.
+    """
+    pennylane_requires = ">=0.17"
```
AlbertMitjans, line 64:
```suggestion
    pennylane_requires = ">=0.18"
```

pennylane_forest/qc.py
```diff
@@ -0,0 +1,296 @@
...
+            raise ValueError(
+                f"Device has a fixed number of {self.num_wires} qubits and "
+                f"cannot be created with {len(wires)} wires."
+            )
+
+        self.wiring = dict(enumerate(self.qc.qubits()))
+        self.active_reset = active_reset
+
+        super().__init__(wires, shots)
+        self.reset()
```
AlbertMitjans, line 121: `self.reset()` is already called in `super().__init__`

pennylane_forest/qc.py
```diff
@@ -0,0 +1,296 @@
...
+
+    @abstractmethod
+    def get_qc(self, device, **kwargs) -> QuantumComputer:
+        """Initializes and returns a pyQuil QuantumComputer that can run quantum programs"""
+
+    @property
+    def circuit_hash(self):
```
AlbertMitjans, line 128: Add docstring

pennylane_forest/device.py
```diff
@@ -191,17 +189,16 @@ def __init__(self, wires, shots=1000):
         super().__init__(wires, shots)
         self.reset()
```
AlbertMitjans, line 190: To avoid the CodeFactor warning:
```
Attribute 'prog' defined outside __init__ (attribute-defined-outside-init)
```

We could do:
```suggestion
        self.prog = Program()
        self._state = None
```

pennylane_forest/qc.py
```diff
@@ -0,0 +1,296 @@
...
+                    # of the corresponding operation
+                    par.append(self._parameter_reference_map[parameter_string])
+                else:
+                    par.append(param)
+
+            prog += self._operation_map[operation.name](*par, *device_wires.labels)
+
+        return prog
+
+    def execute(self, circuit, **kwargs):
```
AlbertMitjans, line 245: Add docstring

pennylane_forest/qc.py
```diff
@@ -0,0 +1,296 @@
...
+        if self.parametric_compilation:
+            self._circuit_hash = circuit.graph.hash
+        return super().execute(circuit, **kwargs)
+
+    def generate_samples(self):
```
AlbertMitjans, line 250: Add docstring

AlbertMitjans (commented):
Amazing job with the refactor!

Added some coments.

pennylane_forest/qc.py
MarquessV: Set the min version of pennylane to `0.18` everywhere.

pennylane_forest/qc.py
MarquessV: Good catch, removed it.

pennylane_forest/qc.py
MarquessV: Added all the requested docstrings - definitely welcome to feedback on them.

pennylane_forest/device.py
MarquessV: I think we have to have a valid `reset()` function since it is part of the `QubitDevice` API, so I don't think we can get around this rule. I added a comment to disable it.

AlbertMitjans: We can keep the `reset()` method, but I was suggesting using:
```python
self.prog = Program()
self._state = None
```
in the `__init__` function instead of calling `reset()`. This way the `prog` and `_state` attributes are defined in the `__init__` function and then CodeFactor won't complain anymore

pennylane_forest/device.py
```diff
@@ -266,6 +266,7 @@ def apply_circuit_operations(self, operations):
 
     def reset(self):
         """Resets the device after the previous run."""
+        # pylint:disable=attribute-defined-outside-init
```
AlbertMitjans, line 269: As I mentioned earlier: if we use
```python
self.prog = Program()
self._state = None
```
instead of calling `reset()` in the `__init__` function, this pylint error won't happen.

AlbertMitjans:
There was a timeout error in the `test_qvm.py:test_qubit_unitary` test. Should we increase the timeout threshold?

pennylane_forest/qpu.py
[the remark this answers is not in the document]
AlbertMitjans: I would leave this as it is for now. We can create a follow-up issue saying that this should be optimised.

pennylane_forest/device.py
MarquessV: Okay, initialized those in `__init__`

pennylane_forest/qpu.py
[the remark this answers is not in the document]
AlbertMitjans: @antalszava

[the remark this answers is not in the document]
antalszava: The QPU device allows operator estimation and uses `measure_observables` from PyQuil under the hood. This feature in the plugin is, however, not compatible with parametric compilation. Therefore, once parametric compilation is turned on, we're falling back to PennyLane's implementation using `generate_samples`.

A major shortcoming hinted at above too is that when doing operator estimation, we're effectively generating samples using PennyLane's pipeline **and** using `measure_observables` - the first of which should not be done.

See https://github.com/PennyLaneAI/pennylane-forest/issues/45 that has just been re-opened.

AlbertMitjans (approved):
Approved!

PR #108, opened 2022-10-25
QPUDevice: Skip generate_samples to avoid executing a circuit twice

AlbertMitjans:
**Context**:
When running the following code, the circuit is executed twice: once by the `Device.generate_samples` method and another one by the `QPUDevice.expval` method, but only the results from the last method are used.
```python
dev = qml.device("forest.qpu", device="4q-qvm", shots=10000, load_qc=False, parametric_compilation=False)

@qml.qnode(dev)
def circuit():
    return qml.counts(qml.PauliZ(0))

print(circuit())
```
**Issue**: https://github.com/PennyLaneAI/pennylane-forest/issues/45
**Description of the change**: Skip `Device.generate_samples` when executing a circuit with `expval` and `parametric_compilation=False`
**Possible drawbacks**: We are overriding the `Device.execute` method, which is not recommended.

### DIFF
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -201,6 +201,4 @@ class QPUDevice(QuantumComputerDevice):
         return super().execute(circuit, **kwargs)
 
     def generate_samples(self):
-        if self._skip_generate_samples:
-            return
-        return super().generate_samples()
+        return None if self._skip_generate_samples else super().generate_samples()
--- a/pennylane_forest/qpu.py
+++ b/pennylane_forest/qpu.py
@@ -22,6 +22,9 @@ Code details
 import warnings
 
 import numpy as np
+from pennylane.measurements import Expectation
+from pennylane.operation import Tensor
+from pennylane.tape import QuantumTape
 from pyquil import get_qc
 from pyquil.api import QuantumComputer
 from pyquil.experiment import SymmetrizationLevel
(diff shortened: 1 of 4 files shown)


pennylane_forest/qpu.py
```diff
@@ -139,11 +141,7 @@ def expval(self, observable, shot_range=None, bin_size=None):
                 pauli_obs = sZ(wire)
 
             # Multi-qubit observable
-            elif (
-                len(device_wires) > 1
-                and isinstance(observable, Tensor)
-                and not self.parametric_compilation
```
antalszava, line 145: Why was this changed? This seems to allow parametric compilation with op estimation, do they work well together in this plugin?

antalszava (commented):
:100: I had one question. Don't forget to add an entry to the changelog. It likely would be required for #107 (and may have been skipped in the past :/).

pennylane_forest/qpu.py
AlbertMitjans: We already check that `self.parametric_compilation` is False at the beginning of this method. This check here did nothing.

AlbertMitjans: I don't think they work well together, so we still disable operator estimation when `parametric_compilation = True`

AlbertMitjans:
> Don't forget to add an entry to the changelog.

Added changelog entry.

> It likely would be required for https://github.com/PennyLaneAI/pennylane-forest/pull/107 (and may have been skipped in the past :/).

Oops my fault. I will tell the people of Rigetti to create a PR adding the changelog.

pennylane_forest/qpu.py
antalszava: Good catch! Thanks for explaining.

antalszava (approved):
Looks good :+1:

PR #110, opened 2022-11-01
Rename PennyLane Forest to PennyLane Rigetti

MarquessV:
~Leaving as draft for now - want to take another pass at this with fresh eyes later.~

Everything looks to be in order. Let me know if I should make any additional changes, or if I can help with the release process at all.


setup.py
```diff
@@ -3,38 +3,38 @@
...
 requirements = ["pyquil>=3.0.0,<4.0.0", "qcs-api-client>=0.20.13<0.22.0", "pennylane>=0.18"]
 
 info = {
-    "name": "PennyLane-Forest",
+    "name": "PennyLane-Rigetti",
     "version": version,
     "maintainer": "Xanadu Inc.",
     "maintainer_email": "software@xanadu.ai",
-    "url": "https://github.com/PennyLaneAI/pennylane-forest",
+    "url": "https://github.com/PennyLaneAI/pennylane-rigetti",
```
MarquessV, line 17: Given all the references here and in the README, it'll take some co-ordination to rename the repo, release the package under a new name, and forward folks from pennylane-forest to pennylane-rigetti on pypi.

josh146: @MarquessV regarding the PyPI redirect - do you have a proposal here? This isn't something we have done in the past

MarquessV: This is what I would propose:
1. Release the package under the _new_ name
2. Create a new release for the package under the _old_ name. This final release should just contain a README directing users to the new package, and a `setup.py` that includes `install_requires = ["pennylane-rigetti"]` so if users do install the package under the old name, they will get redirected to the new one.

This [repo](https://github.com/simonw/pypi-rename) provides a tool to generate the final release under the old name. Maybe start with that, and customize it as needed, or just use [their template](https://github.com/simonw/pypi-rename/tree/main/%7B%7Bcookiecutter.old_package_name%7D%7D) as a base to work off of. 

I would also recommend attempting all of this on [TestPyPI](https://test.pypi.org/) first, that way we can see what things will look like before going live with it.

Jaybsoni (approved):
Looks good to go 👍🏼

PR #112, opened 2022-11-07
Old 0.24 changelog has the wrong PR linked

timmysilv:
as title says. was presumably a copy-paste from a similar change in the qulacs plugin. Note that this targets the pre-release branch from #111


antalszava (approved):
Great catch :+1:

PR #121, opened 2023-02-08
Bump python version

AlbertMitjans:
- Copy `.readthedocs.yaml` configuration file from PL to avoid using `python3.7` to build the docs.
- Update `tests.yml` workflow to stop using `python3.7` in `integration-tests`.
- Run `black`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,14 +5,13 @@ on:
       - master
   pull_request:
 
-
 jobs:
   tests:
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        python-version: [3.8, '3.10']
+        python-version: ["3.8", "3.10"]
 
     steps:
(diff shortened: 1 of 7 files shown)


MarquessV (approved):
I don't think my reviews hold any weight here, but looks good!

timmysilv:
is this a formal end for 3.7 support for this plugin? if so, can you remove it from `setup.py` and the README (bump to 3.8 there)? otherwise lgtm!

AlbertMitjans:
> if so, can you remove it from setup.py and the README (bump to 3.8 there)?

Good catch! Done 👍

AlbertMitjans:
> is this a formal end for 3.7 support for this plugin?
Yes!

timmysilv (approved):
also now realizing we should add a changelog entry to formalize the end of 3.7 support... please add that and 🚢 🚢 !

PR #122, opened 2023-02-09
Removed in-place inversion

mudit2812:
Replaced all instances of in-place inversion with `qml.adjoint()`


tests/test_converter.py
```diff
@@ -18,116 +18,152 @@ class TestProgramConverter:
         [
             (g.I(0), qml.Identity(wires=[0])),
             (g.H(0), qml.Hadamard(0)),
-            (g.H(0).dagger(), qml.Hadamard(0).inv()),
-            (g.H(0).dagger().dagger(), qml.Hadamard(0).inv().inv()),
+            (g.H(0).dagger(), qml.adjoint(qml.Hadamard(0))),
+            (g.H(0).dagger().dagger(), qml.Hadamard(0)),
```
AlbertMitjans, line 22: Would be nice to be consistent with the changes
```suggestion
            (g.H(0).dagger().dagger(), qml.adjoint(qml.adjoint(qml.Hadamard(0)))),
```

AlbertMitjans: Same thing for all operations

mudit2812: I did that before and the tests failed. If a pyquil gate is inverted multiple times, the code in `pennylane_rigetti/converter.py:_is_inverted` only checks whether a gate is inverted or not by counting the number of times a gate has been inverted and checking if the count is odd or even.

PR #124, opened 2023-02-16
Fix default wire mapping and QPU program compilation and add support for parallel execution

MarquessV:
This addresses #123 in a 3 ways:

1. Default wire mapping behavior

By default, we were using `[0, n)` as the wire numbers for an n-qubit device, but qubits on a real QPU aren't always addressed in this way. Instead, we let `pyQuil` tell us what the qubits on the device are and build the wire map from that. By default, the user wire labels will map to a device wire label 1-to-1, that way users who want to address them, can do so using the values directly from the our QPU topology metrics and diagrams.

2. Qubit measurement

Measurements will now use the wire mapping to generate `MEASURE` statements for every Qubit to readout instead of assuming `[0, n)`. Like (1), this is important for real QPUs, which don't always address qubits sequentially.

3. Use `protoquil` for program compilation targeting QPUs

By default a program compiled by `quilc` will use the full `quil` spec to generate native `quil`. However, not everything in the `quil` spec is supported by our QPUs, so `quilc` also supports building a program using `protoquil`, a subset of `quil` that is compatible with a QPU.

Our QPU devices now set the `protoquil` flag when compiling to avoid generating a program that might not run on a real QPU.

This also adds support for parallel execution using threadpools to run jobs in parallel. This works for both the QVM and QPU device and can greatly improve execution speed.


pennylane_rigetti/qc.py
```diff
@@ -180,10 +178,8 @@ def define_wire_map(self, wires):
 
     def apply(self, operations, **kwargs):
         """Applies the given quantum operations."""
-        prag = Program(Pragma("INITIAL_REWIRING", ['"PARTIAL"']))
```
MarquessV, line 183: By default, the compiler will choose `NAIVE` if the program fits the topological constraints of the QPU, and `PARTIAL` otherwise. I've removed this so that we can lean on that behavior instead of overwriting it for every program.

pennylane_rigetti/qc.py
```diff
@@ -192,13 +188,14 @@ def apply(self, operations, **kwargs):
...
-            self.prog.inst(MEASURE(q, ro[i]))
-
         self.prog.wrap_in_numshots_loop(self.shots)
+        # Measure every qubit used by the program into a readout register.
+        # Devices don't always have sequentially adressed qubits, so
+        # we use a normalized value to index them into the readout register.
+        used_qubits = self.prog.get_qubits(indices=True)
+        ro = self.prog.declare("ro", "BIT", len(used_qubits))
+        for i, qubit in enumerate(used_qubits):
+            self.prog += MEASURE(qubit, ro[i])
```
MarquessV, line 198: This is a large departure from how the plugin was originally written so I want to draw extra attention to this and solicit some extra feedback and clarification.

Is there any reason we should still be measuring every qubit here?

The `ro` register we declare here is ultimately what gets returned by `generate_samples`. What is the expected shape of the matrix `generated_samples` returns? Should it relate to the number of qubits a circuit actually uses, or maybe the number of qubits on a circuit that are measured?

AlbertMitjans: In PennyLane, the shape of the output of a circuit depends on the number of qubits **measured**, and not the number of qubits **used** in a circuit. Here is a dummy example:
```python
import pennylane as qml

dev = qml.device("default.qubit", wires=4, shots=1000)


@qml.qnode(dev)
def circuit():
    qml.PauliX(0)
    qml.PauliX(1)
    qml.PauliX(2)
    return qml.probs(op=qml.PauliZ(0) @ qml.PauliZ(1))


print(qml.math.shape(circuit()))
```
Which returns: `(4,)` instead of `(8,)`.

AlbertMitjans: `generate_samples` uses all the qubits of the **device**! Thus if a device has 10 qubits, and the circuit executed uses 5 qubits and measures 3, we still return 10 qubits in `generate_samples` (I know, it is not very optimal).

MarquessV: Gotcha, I'll fit the data into a register of that size then. Based on the old code, it looks like it's expected that each row of samples should be ordered from the lowest valued qubit label to the highest valued qubit label, so I went with that pattern. Let me know if it should be something different though!

pennylane_rigetti/qc.py
```diff
@@ -45,15 +46,17 @@ class QuantumComputerDevice(RigettiDevice, ABC):
         device (str): the name of the device to initialise.
         shots (int): number of circuit evaluations/random samples used
             to estimate expectation values of observables.
-         wires (Iterable[Number, str]): Iterable that contains unique labels for the
+        wires (Iterable[Number, str]): Iterable that contains unique labels for the
             qubits as numbers or strings (i.e., ``['q1', ..., 'qN']``).
             The number of labels must match the number of qubits accessible on the backend.
-            If not provided, qubits are addressed as consecutive integers ``[0, 1, ...]``, and their number
-            is inferred from the backend.
+            If not provided, qubits are addressed by the backend.
```
trbromley, line 52: To make sure I understand, does this mean that the user can pass _any_ wires and they'll be mapped to the device wires, or that the user must use the appropriate wires for the device?

MarquessV:
Per offline discussion, closing this in favorite of separate PRs for the wire mapping fix and the parallel processing improvement.

PR #120, opened 2023-02-07
Fix missing comma

genos:
In #119, we noted that the missing comma in `qcs-api-client>=0.20.13<0.22.0` in `setup.py` was causing issues. With the changes here, the reported problem disappears:

```python
>>> import pennylane as qml
>>> qml.device("rigetti.qpu", device="Aspen-M-2")
<QPUDevice device (wires=80, shots=1000) at 0x132e949d0>
```

What's more, with `pip install -e '.[test]'` in a virtual environment, all unit tests pass (with some warnings).

PR #125, opened 2023-02-21
update changelog before plugin version bumps

timmysilv:
- fix QubitStateVector initialization in test to be valid now the PL checks for it
- moved the bug component of #120 to the bug fixes section
- added @mudit2812 as a contributor


rmoyard (approved):
Thanks :100:

PR #129, opened 2023-04-11
update tests to use np.allclose and to not use qml.map

timmysilv:
Two things made tests fails:
1. tests were using `qml.map` which creates QNodeCollections. Those are deprecated, and don't work with the new return system
2. `assertAllAlmostEqual` has custom logic if the first arg (the qnode result) is a tuple, which they now are with the new return system. Unfortunately, the custom logic is not what we want in the regular case

Solutions:
1. Stop using QNodeCollection. Sometimes I just return the same observable 6 times, sometimes we don't even need a collection
2. Just use `np.allclose` (renaming `delta` to `atol`) in those cases because it's what we want anyway

### DIFF
--- a/tests/test_pyqvm.py
+++ b/tests/test_pyqvm.py
@@ -39,7 +39,7 @@ class TestPyQVMBasic(BaseTest):
         res = circuit()
 
         # below are the analytic expectation values for this circuit (trace should always be 1)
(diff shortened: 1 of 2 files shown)


timmysilv:
[sc-35013]

rmoyard (approved):
Thanks @timmysilv it looks good to me :+1:

lillian542 (approved):
Looks good to me!

PR #133, opened 2023-06-26
Switch to GA4 tag

doctorperceptron:
Google is deprecating Universal Analytics on July 1st. This PR adds the newer GA4 tag.


timmysilv (approved):
thanks @doctorperceptron! ~are you making this change across all plugins?~
saw the SF PR so I'm guessing yes. thank you very much!

PR #139, opened 2023-08-23
put back pennylane version

timmysilv:
leaving as draft until PL 0.32 is released

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-pennylane @ git+https://github.com/PennyLaneAI/pennylane.git
 numpy>=1.21,<1.24
 networkx>=2.5,<3.0
+pennylane>=0.32
 pyquil>=3.0.0,<4.0.0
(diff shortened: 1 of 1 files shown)


trbromley (commented):
Will this change fix https://github.com/PennyLaneAI/plugin-test-matrix/actions/runs/5970113161/job/16197169878?

timmysilv:
well yes... but it can't be merged until 0.32 is released. if you like, we can change master to pin to the PL RC branch so at least this CI turns red, then change it to this (>=0.32) before we release on Monday

mudit2812:
> well yes... but it can't be merged until 0.32 is released. if you like, we can change master to pin to the PL RC branch so at least this CI turns red, then change it to this (>=0.32) before we release on Monday

I'm okay with this. The version bumps will be needed for a few other plugins as well I think.

lillian542 (approved):

PR #142, opened 2023-09-21
unpin numpy and fix BasisState test

timmysilv:
this BasisState test fails because PL now uses the inputs as numpy array indices, so they should be integers. We can cast to ints on the PL side, but this should do for now. If people feel strongly about this casting, I can do it at the source (should be pretty straightforward)

### DIFF
--- a/tests/test_gradients.py
+++ b/tests/test_gradients.py
@@ -14,7 +14,7 @@ def test_simulator_qvm_default_agree(tol, qvm, compiler):
     dev2 = qml.device("rigetti.wavefunction", wires=w)
     dev3 = qml.device("rigetti.qvm", device="9q-square-qvm", shots=5000)
 
-    in_state = np.zeros([w], requires_grad=False)
+    in_state = np.zeros([w], dtype=np.int64, requires_grad=False)
     in_state[0] = 1
     in_state[1] = 1
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-numpy>=1.21,<1.24
+numpy>=1.21
 networkx>=2.5,<3.0
 pennylane>=0.32
 pyquil>=3.0.0,<4.0.0


mudit2812 (approved):
I don't have a strong opinion, but maybe wait for someone who might before merging

timmysilv:
cc @trbromley

tests/test_gradients.py
```diff
@@ -14,7 +14,7 @@ def test_simulator_qvm_default_agree(tol, qvm, compiler):
     dev2 = qml.device("rigetti.wavefunction", wires=w)
     dev3 = qml.device("rigetti.qvm", device="9q-square-qvm", shots=5000)
 
-    in_state = np.zeros([w], requires_grad=False)
+    in_state = np.zeros([w], dtype=np.int64, requires_grad=False)
```
trbromley, line 17: Why does the unpinning lead to this issue? Or is this unrelated to the pinning?

trbromley: Thanks! I have a preference to cast to ints internally if this is something we did before.

timmysilv: [what we did before](https://github.com/PennyLaneAI/pennylane/blob/b7279db478fa144d6da91c1177e2845a21a28551/pennylane/devices/default_qubit_legacy.py#L793-L797) is completely different from [what we do now](https://github.com/PennyLaneAI/pennylane/blob/b7279db478fa144d6da91c1177e2845a21a28551/pennylane/ops/qubit/state_preparation.py#L118-L120), but floats would be accepted in the old way.

I'd still like to make this change here because `in_state` _should_ be integers. I'll cast to int in PennyLane itself as well, but I think this change makes sense regardless of the bug.

trbromley: > I'll cast to int in PennyLane itself as well

Thanks @timmysilv!

trbromley: This might already be causing an issue in the dev QML builds:
https://github.com/PennyLaneAI/qml/actions/runs/6306272465/job/17121149404#step:18:1593

This bumps up the priority of fixing the casting to help unblock the dev builds.

timmysilv: full fix: PennyLaneAI/pennylane#4632

trbromley (approved):

PR #158, opened 2024-05-13
Install latest PennyLane for CI

astralcai:


astralcai:
[sc-63171]

.github/workflows/tests.yml
```diff
@@ -50,7 +51,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.9", "3.11"]
+        python-version: ["3.9", "3.10", "3.11"]
```
dwierichs, line 54: Why was 3.10 not included here, previously? :thinking:

.github/workflows/tests.yml
```diff
@@ -75,6 +76,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install '.[test]'
+          pip install --upgrade git+https://github.com/PennyLaneAI/pennylane.git#egg=pennylane
```
dwierichs, line 79: Should we keep the ordering identical between tests and integration-tests :thinking:

dwierichs (approved):
LGTM, just two tiny questions mostly out of curiousity :)

.github/workflows/tests.yml
astralcai: So it was previously "3.8" and "3.10", and then it got changed to "3.9" and "3.11". I don't see why "3.10" shouldn't be included so I added it back in.

.github/workflows/tests.yml
astralcai, line 79:
```suggestion
          pip install --upgrade git+https://github.com/PennyLaneAI/pennylane.git#egg=pennylane
          pip install '.[test]'
```

PR #174, opened 2025-01-31
Quick fix return type

JerryChen97:
Since it's qutie minimum we can still try to apply the changes to see if latest fail will disappear.

### DIFF
--- a/pennylane_rigetti/qpu.py
+++ b/pennylane_rigetti/qpu.py
@@ -23,7 +23,7 @@ Code details
 import warnings
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


andrijapau:
I don't think this is necessary since this fix would mean we would have to release a new version of Rigetti (which is something we are not currently maintaining)

JerryChen97:
> I don't think this is necessary since this fix would mean we would have to release a new version of Rigetti (which is something we are not currently maintaining)

Oh yeah this is right. let me close it

